PR #29, opened 2014-06-05
cluster-ha-singleton: service should depend on SchedulerBean

sebek64:
Although this small example works even without the fix, a bigger deployment
can fail due to a race condition - SchedulerBean might be activated after
the HASingletonService is started.


n1hility:
@pferraro wdyt?

pferraro:
Oye.  I guess I missed some pings...
We definitely don't want to recommend that people implementing singleton services use such a dependency.  This use case should be obsoleted by singleton deployments as of WF10.

PR #58, opened 2015-03-04
Fix typo: Should be @Model instead of @Member

kmix:
@Model annotation adds @Named & @RequestScoped to a class.


ctomc:
thank you for the fix

PR #60, opened 2015-04-14
Updated destinations creation process description

mgorgon:
hornetq-jms.xml is no longer used in this project - destinations are created via annotations


ctomc:
Thank you

PR #61, opened 2015-05-03
Wfly 4571

pgier:
Update quickstarts to use consolidated BOMs
Update carmart quickstarts to use Infinispan 7

### DIFF
--- a/carmart-tx/pom.xml
+++ b/carmart-tx/pom.xml
@@ -23,7 +23,7 @@
     <artifactId>wildfly-carmart-tx</artifactId>
     <version>9.0.0-SNAPSHOT</version>
     <packaging>war</packaging>
-    <name>Transactional CarMart</name>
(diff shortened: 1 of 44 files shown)


carmart-tx/pom.xml
```diff
@@ -111,6 +111,12 @@
             <artifactId>infinispan-core</artifactId>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.infinispan</groupId>
+            <artifactId>infinispan-commons</artifactId>
+            <version>7.1.1.Final</version>
```
ctomc: couldn't this be replaced by ${version.org.infinispan} which should get imported from bom.

ctomc: nevermind that wont work.

pgier: This version setting should just be temporary until we can either add infinispan-commons into the bom, or refactor the quickstart so that it doesn't need this jar.

PR #66, opened 2015-08-30
helloworld-ssl example to demonstrate ssl configuration over wildfly.

girirajsharma:
This is first and simplest example (wildfly ssl configuration) in the wildlfy-ssl series. It will be followed by 
1. Mutual (two way) ssl authentication for clients over wildfly
2. Securing war (client) apps over wildlfy
3. Some other security tweaks/mechanisms with wildfly.


ctomc:
Thank you!

PR #67, opened 2015-09-03
helloworld-client-ssl example to show mutual ssl auth over wildfly.

girirajsharma:
This example demonstrates mutual client ssl authentication over wildfly or JBoss application server via configuring keystore, truststore and https listener. Keys and certificates for configuring wildfly standalone.conf have been generated using openSSL keytool.

### DIFF
--- a/helloworld-client-ssl/README.md
+++ b/helloworld-client-ssl/README.md
@@ -1,0 +1,302 @@
+helloworld-client-ssl: Wildfly mutual SSL(two-way) configuration example
+=======================================================================
+Author: Giriraj Sharma
+
+Level: Intermediate
+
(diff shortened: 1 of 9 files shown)


ctomc:
First thank you for good quickstart. great work.
I think would be better if instead or in addition of using openssl, manual should tell how to use java's keytool. This way you get cross platform manual and make sure there are no issues with loading keystores that jdk cannot digest.

girirajsharma:
@ctomc Thanks.
+1. Agreed.
I will do the required changes i.e., providing manual for both Java Keytool and openSSL and will update/squash the PR soon.

girirajsharma:
Updated

ctomc:
Thank you!

PR #70, opened 2015-09-08
JDF-841 - Synchronize Wildfly and EAP Quickstarts

rafabene:
This PR merges the changes from EAP (7.0) to WildFly


app-client/pom.xml
```diff
@@ -89,27 +82,11 @@
...
-            </dependency>
         </dependencies>
     </dependencyManagement>
 
     <build>
         <plugins>
-            <!-- Enforce Java X -->
-            <plugin>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <version>${version.compiler.plugin}</version>
```
ctomc: thing is that if we don't specify version of compiler plugin it will default to either 2.3.2 which doesn't properly handle annotation processors or 3.2/3.3 which is broken on recompile.
one option would be use jboss-parent which defines all plugin versions.

ctomc:
I am going to merge this as it is huge as it is already, all other changes discussed here should be done in separate PR.

ctomc:
Uff, just realized this pr is for 9.x branch, it should be for 10.x not 9.
EAP7 is based on 10 not 9

PR #71, opened 2015-09-08
Revert "JDF-841 - Synchronize Wildfly and EAP Quickstarts"

ctomc:
Reverts wildfly/quickstart#70

this needs to go to 10.x branch not 9.x


ctomc:
@rafabene reverting this as it needs to be in 10.x

PR #74, opened 2015-09-16
WFLY-5250 - fix jts quickstart validation error on jboss-ejb3.xml

robstryker:


rafabene:
i think this should be fixed on EAP version too.

robstryker:
@rafabene  This is already fixed in EAP 7 quickstarts.

rafabene:
The file content on EAP 7 QS is different: https://github.com/jboss-developer/jboss-eap-quickstarts/blob/7.0.x-develop/jts/application-component-2/src/main/resources/META-INF/jboss-ejb3.xml

PR #75, opened 2015-09-16
WFLY-5244 - comment out property tag in arquillian.xml so xml validates properly

robstryker:


rafabene:
This PR looks ok and conforms to EAP version: https://github.com/jboss-developer/jboss-eap-quickstarts/blob/7.0.x-develop/bean-validation/src/test/resources/arquillian.xml

PR #76, opened 2015-09-16
WFLY-5247 - unify wf ejb-multi-server quickstart with eap7 quickstart

robstryker:
This patch unifies the ejb-multi-server quickstart's descriptor with that of the eap7.  

I am not entirely sure this patch is needed, or whether it is better to have no jboss-ejb3.xml file, or whether the one in EAP is a good choice. But I trust that the EAP quickstarts have done a good job and this file seems to be a better addition to help guide the users.


rafabene:
This file was wrongly removed and was returned as part of https://github.com/wildfly/quickstart/commit/d77feccae9edb1d5f275c1764c81147a1556ae82#diff-124fd1ca9ddc5dc3669ecbe106c98076

PR #79, opened 2015-10-29
WebSocket Client

monospacesoftware:
quickstart needs a websocket client example.  My changes expand on wildfly-helloworld-websocket by adding a @ClientEndpoint that connects to the public ws://echo.websocket.org echo server, and passes messages using CDI events.  (So it's a good example of CDI events also.)


ctomc:
this needs to be for 10.x branch not 9.x, 10 is already on JDK8

monospacesoftware:
I will resubmit on 10.x with some cleanups.

PR #80, opened 2015-11-02
Cleanup typos and maven inconsistencies

AlexBischof:
Hi,

this PR fixes typos in **README.MD** where to old urls referring to jboss instead of wildfly. I also renamed all submodules (e.g. cdi-decorator) and dependencies which still referring to jboss instead of wildfly.

Kind regards
Alex

### DIFF
--- a/batch-processing/README.md
+++ b/batch-processing/README.md
@@ -60,7 +60,7 @@ Build and Deploy the Quickstart
 Access the application
 ---------------------
 
-Access the running application in a browser at the following URL:  <http://localhost:8080/jboss-batch-processing/>
+Access the running application in a browser at the following URL:  <http://localhost:8080/wildfly-batch-processing/>
 
 You're presented with a simple form that allows you to generate sample files to be imported. 
 
--- a/bmt/README.md
+++ b/bmt/README.md
@@ -66,7 +66,7 @@ Build and Deploy the Quickstart
 Access the application 
 ---------------------
 
-The application will be running at the following URL: <http://localhost:8080/jboss-bmt/>.
+The application will be running at the following URL: <http://localhost:8080/wildfly-bmt/>.
 
 You will be presented with a simple form for adding key/value pairs and a checkbox to indicate whether the updates should be executed using an unmanaged component. Effectively this will run the transaction and JPA updates in the servlet, not session beans. If the box is checked then the updates will be executed within a session bean method.
 
--- a/cdi-alternative/README.md
+++ b/cdi-alternative/README.md
@@ -61,7 +61,7 @@ Build and Deploy the Quickstart
 Access the application
 ---------------------
 
-The application will be running at the following URL: <http://localhost:8080/jboss-cdi-alternative>.
+The application will be running at the following URL: <http://localhost:8080/wildfly-cdi-alternative>.
 
 You can specify alternative versions of the bean in the WEB-INF/beans.xml file by doing one of the following:
 
--- a/cdi-decorator/README.md
+++ b/cdi-decorator/README.md
@@ -57,7 +57,7 @@ Build and Deploy the Quickstart
 Access the application
 ---------------------
 
-The application will be running at the following URL: <http://localhost:8080/jboss-cdi-decorator>.
+The application will be running at the following URL: <http://localhost:8080/wildfly-cdi-decorator>.
 
 You can specify decorator of the bean in the WEB-INF/beans.xml file by doing one of the following:
 
--- a/cdi-injection/README.md
+++ b/cdi-injection/README.md
@@ -52,7 +52,7 @@ Build and Deploy the Quickstart
 Access the application 
 ---------------------
 
-The application will be running at the following URL <http://localhost:8080/jboss-cdi-injection/>.
+The application will be running at the following URL <http://localhost:8080/wildfly-cdi-injection/>.
 
 
 Undeploy the Archive
--- a/cdi-interceptors/README.md
+++ b/cdi-interceptors/README.md
@@ -60,7 +60,7 @@ Build and Deploy the Quickstart
 Access the application 
 ---------------------
 
-The application will be running at the following URL: <http://localhost:8080/jboss-cdi-interceptors>.
+The application will be running at the following URL: <http://localhost:8080/wildfly-cdi-interceptors>.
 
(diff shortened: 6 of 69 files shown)


rhusar:
@sgilda I am just wondering if it would not be better to be a bit more agnostic to the target runtime and avoid frequently renaming jboss/wildfly/eap; for instance context paths not need to change from jboss to wildfly, but could be simply "/quickstart-xyz".

ctomc:
this pr needs to be rebased

rhusar:
@ctomc Well, thats what I was going to say too before realizing there is probably better way than changing things back and forth, thus comment ^

sgilda:
@rhusar , @ctomc , @AlexBischof : 

I believe the original idea for the `jboss-` prefix in the context-path for the quickstarts was to avoid clobbering another (non-JBoss) deployed applications. 

If we leave off the prefix, a customer could conceivably deploy one of our quickstarts over a real application or another example. I think we can assume customers do not use the 'jboss-' prefix in the context-path for their own application projects. 

I am copying @pmuir and @rafabene for their feedback as I'm sure their memories may be better than mine.

ctomc:
given that this are quickstarts for people to try our server I doubt they have any thing else deployed on that same server.

rafabene:
The initial  use for the prefix- was to avoid collision with existing applications. But I also agree that it's being a pain to keep wildfly- and jboss-. My proposal is that we change it to redhat- for the next EAP release so it can easily be used by wildfly and eap quickstarts.

sgilda:
@rafabene : That sounds like a great plan for the next release! 

Does this work for everyone else?

rhusar:
:-1: 
I don't see why the name of the company should be involved in e.g. the web context. As you say the point of prefix is to differentiate deployments, I would still say /quickstart-xyz would do the job and solve the humdrum rewriting wildfly/eap.

sgilda:
@rhusar : What @rafabene is saying is that we replace both the 'wildfly-' and 'jboss-' prefixes with 'redhat-'. That way it is the same prefix used in the WildFly and JBoss EAP quickstarts and they don't ever need to be changed again. They are Red Hat quickstarts, so the 'redhat-' prefix distinguishes them from deployments of quickstart or examples from other organizations.

Does this explain it better?

ctomc:
Well if argument is that they are "Red Hat" quickstarts, which no one can object, than it is also true that they are "jboss" quickstarts, with "jboss" as in jboss.org community not as in jboss EAP or Jboss AS
so having either jboss or redhat as prefix makes sense.
In any case I am not a fan of any prefix for that matter.

rafabene:
/quickstart-xyz , /jboss- or /redhat- . Any of them are ok as long as we still using a prefix and we avoid rewriting wildfly/eap.

rhusar:
My perception is that  we have been trying hard to make sure users are able to differentiate community from the product. For instance, if you look at http://wildfly.org/ you can barely see references to Red Hat or JBoss. Those are names we want people to have associated with commercial offering. So as long as the prefixes are not /jboss- or /redhat- we are following the vision in my original comment https://github.com/wildfly/quickstart/pull/80#issuecomment-172903645 or otherwise we would need to keep wildfly here and jboss/redhat for the product.

sgilda:
I still think 'quickstart-' is too generic and too risky. It does not branch the quickstart as ours and differentiate it from other examples.

rhusar:
Branch? Meaning 'brand'? Well, you will still see a lot of "wildfly" in package names and references where it cannot be changed to EAP-branding.

ctomc:
why does it matter how it differentiate? name jboss / wildfly is used in source packages, in maven coordinates in dependencies etc etc.

rhusar:
Exactly, that's why I am opposing introducing new term such as redhat-.

rafabene:
This discussion seems legit from the perspective of creating a
differentiation between the project and the product. But shouldn't we
consider the impact  on the maintenance of both quickstarts? We frequently
send fixes to from downstream to upstream and vice versa. This kind of
change could impact these fixes since it creates potential merging
conflicts.

On Wed, Jan 20, 2016 at 11:23 AM, Radoslav Husar <notifications@github.com>
wrote:

> Exactly, that's why I am opposing introducing new term such as redhat-.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/wildfly/quickstart/pull/80#issuecomment-173259341>.
>



-- 
*Rafael Benevides | Senior Software Engineer*
JBoss Developer Materials lead
M: +1-919-592-6255



Better technology. Faster innovation. Powered by community collaboration.
See how it works at www.redhat.com

rhusar:
> But shouldn't we consider the impact  on the maintenance of both quickstarts? 

That is exactly the reason why I started this discussion (because this PR changes that to WF which is not suitable for EAP branch of this). Using terms that are suitable for both WF/EAP is what I am proposing. The differentiation is mainly done around dependencies, readme-s, JDF repository, etc.

rafabene:
So it seems that /quickstart-xyz is the better approach to satisfy both worlds (project and product). As I've said previously, I don't have any objections to use /quickstart-xyz

rhusar:
:+1:

sgilda:
I would still like to hear from @pmuir on this.

sgilda:
I am not a fan of the 'quickstart-' prefix. It is too generic and could potentially conflict with other product quickstarts.
 
Regardless of the prefix we use, we still have the issue with product names in the README.md and other source files, plus differences in artifacts between them. Early on, we tried to keep the community and product quickstarts as close to the same as possible, but there were still a lot of maintenance issues. After WildFly, the product team requested changes in the JBoss EAP quickstarts to use the correct product name branding and they began to diverge more. 

Be aware that I also have a JIRA [1] to change source file package names from 'org.jboss.as.quickstarts' to 'org.jboss.eap.quickstarts'. I rejected the JIRA, but I'm guessing it will come back to haunt me in a future release. This will further complicate things.

[1] https://issues.jboss.org/browse/JBEAP-2924

rhusar:
If anything, it would sound to me the package names should mirror WildFly/EAP package naming -- which is indeed org.wildfly -- and its not going anywhere. Commented on the Jira too for what its worth :-)

ctomc:
I am closing this PRa as it is out of date

PR #95, opened 2016-02-07
Cluster mdb - hornetq -> activemq

dmejer:
wildfly 10.x uses ActiveMQ-artemis instead of HornetQ. I've changed hornetq-clustering to cluster-mdb and change quickstart to activemq. Now this quickstart is working for both standalone and domain modes.


ctomc:
thank you

PR #115, opened 2016-11-24
deployment url prefix correction.

soul2zimate:
deployment prefix in URL is still jboss, it should be wildfly


ctomc:
Prefix is about to change again. So I don't want to merge this now.
we are changing all prefixes to qs- for wildfly and eap QS so it will be easier to maintain two repos.

PR #122, opened 2017-01-13
Update helloworld-jms & ejb-remote quickstarts

jmesnil:
* add *temporary* dependency to org.jboss.modules:jboss-modules
  until REM3-244 and ELY-872 are fixed
* add wildfly-config.xml to the client application so that they
  can authenticate to the server.


jmesnil:
@darranl could you approve please? I'm not sure if the presence of wildfly-config.xml is required but I was not able to auth the apps without its presence.

dudaerich:
I think the wildfy-config.xml shouldn't be required until you need to change default configuration. It breaks backward compatibility. The same client's code which worked with 7.0 should work also with 7.1.

darranl:
The wildfly-config.xml shoud also be temporary - we want a default config to be discovered for reasonable out of the box experience.

ctomc:
@darranl @jmesnil is this safe to merge?

ctomc:
Cherry-picked manually, tnx

PR #126, opened 2017-03-02
Use annotations for client interceptor configuration examples

dmlloyd:


sgilda:
@dmlloyd : The README.md file for this quickstart states:

This quickstart uses the ServiceLoader mechanism for registering the EJB client interceptor and places the `META-INF/services/org.jboss.ejb.client.EJBClientInterceptor` in the classpath, with the following content:

        # EJB client interceptor(s) that will be added to the end of the interceptor chain during an invocation
        # on EJB. If these interceptors are to be added at a specific position, other than last, then use the
        # programmatic API in the application to register it explicitly to the EJBClientContext

        org.jboss.as.quickstarts.ejb_security_interceptors.ClientSecurityInterceptor

Does this or anything else in the README.md file need to be updated?

PR #137, opened 2017-03-09
JBEAP-9405: resteasy-jaxrs-client - Modify POM file to set the skip e…

sgilda:
…lement to false so that the tests are not skipped


ctomc:
it is failing tests on CI, I need to investigate what is going on.

ctomc:
This is skipped as it is not really a proper test. As it depends on another QS to be deployed.
Code in this QS needs to be modified, to public static void main class, and instructions how to use that.

PR #152, opened 2017-04-18
[WFLY-8592] adding xts dependency to wsat-simple maven war plugin

ochaloup:
Maven war plugin should define dependency on xts module for packaged war file could be deployed for configuration `standalone-xts.xml` without errors.

https://issues.jboss.org/browse/WFLY-8592
https://github.com/jbossas/eap-quickstarts/pull/34

### DIFF
--- a/wsat-simple/pom.xml
+++ b/wsat-simple/pom.xml
@@ -92,5 +92,19 @@
     <build>
         <!-- Set the name of the WAR, used as the context root when the app is deployed. -->
         <finalName>${project.artifactId}</finalName>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-war-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifest/>
+                        <manifestEntries>
+                            <Dependencies>org.jboss.xts</Dependencies>
+                        </manifestEntries>
+                    </archive>
+                </configuration>
+            </plugin>
+        </plugins>
     </build>
 </project>


ochaloup:
hi @ctomc , can you please take a look on this PR if it is ok to add dependency in quickstarts in this way? Thanks.

ctomc:
@ochaloup is that module public? I don't mind setting dependency like this if modules are public and we do document why it is like that.

wsat-simple/pom.xml
```diff
@@ -92,5 +92,19 @@
...
         <finalName>${project.artifactId}</finalName>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-war-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifest/>
+                        <manifestEntries>
+                          <Dependencies>org.jboss.xts</Dependencies>
```
rhusar: You want to run the code formatter, wrong indent here

ochaloup: it should be addressed, thank you for pointing that out

ochaloup:
@ctomc the module is public, that's ok

the need of using the dependency on the xts module is not documented clearly, for me being honest. Mostly the quickstart shows that.

If user uses Narayana compensating annotations there is already xts processor which adds xts module dependency automatically for the deployment. I'm now thinking to enhance the functionality to cover the case of this quickstart too (this quickstart only implements an interface and adds handler in xml descriptor). But I don't know when I manage to do so.

ctomc:
@ochaloup it would be great to improve the docs & subsystem but it will be fine for now.

PR #170, opened 2017-06-21
[WFLY-8976] Add new quickstart for message-clustering-singleton

fl4via:
Quickstart for message-clustering-singleton.

WFLY Jira: https://issues.jboss.org/browse/WFLY-8976
EAP7 Jira: https://issues.jboss.org/browse/EAP7-714

### DIFF
--- a/messaging-clustering-singleton/README.md
+++ b/messaging-clustering-singleton/README.md
@@ -1,0 +1,374 @@
+# messaging-clustering: Messaging Example that Demonstrates Clustering
+
+Author: Flavia Rainone, Jess Sightler  
+Level: Advanced  
+Technologies: JMS, MDB, Clustering  
+Summary: The `messaging-clustering-singleton` quickstart uses a JMS topic and a queue to demonstrate clustering using ActiveMQMessaging with MDB singleton configuration (only one node in the cluster will be active)
+Target Product: ${product.name}  
+Source: <${github.repo.url}>  
+
+## What is it?
+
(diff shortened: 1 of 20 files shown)


fl4via:
@ctomc  I'm not sure about one thing: I added the delivery-group feature in this quickstart as well, as it is a more advanced feature of mdb deliveries in clusters (and I pointed in the tutorial that it can be used without cluster, btw). I have two questions:
1. Should there be a quickstart only for delivery-groups?
2. Should I include this right now, as I found that the feature was recently broken?
  See https://issues.jboss.org/browse/WFLY-8977 if you need the Jira.

fl4via:
I have another question: this time I saw the guidelines for contributing to this project, and I noticed that it requires artifactId to start with wildfly. So, instead of messaging-clustering-singleton, I should use wildfly-messaging-clustering-singleton, is that it?
I'm double-checking before updating, because the name will be big and this name is used in the README file all over. And, also, because the quickstart I based on, mesaging-clustering, does not have an artifactId name starting with wildfly. Please, let me know if I need to fix this.

fl4via:
@ctomc I'm assuming we are waiting for the resolution of WFLY-8977 as this PR has not been merged. I'll update this PR as soon as WFLY-8977 is solved, the plan is to do so this in the next few days.

ctomc:
I am mostly waiting for downstream PR...

fl4via:
@ctomc Aha! I thought there would be need to change the quickstart due to some feedback, and that's why I held back the upstream PR.
Okay, I'm working on WFLY-8977 right now, and then I'll make the downstream PR right away. Thanks!

fl4via:
EAP PR: https://github.com/jboss-developer/jboss-eap-quickstarts/pull/2069

PR #176, opened 2017-07-12
WFLY-9079 maven-checkstyle-plugin should not run on generated sources

rhusar:
https://issues.jboss.org/browse/WFLY-9079


rhusar:
See the Jira for details and downstream PR for unification. Also fixes build and CI as well on upstream.

PR #185, opened 2017-07-21
JBEAP-12668: Remove trailing spaces from the Spring-based arquillian.…

sgilda:
…xml files to prevent error on import into JBDS


ppalaga:
Note that the Jira in the commit message should have been JBEAP-12268 rather than JBEAP-12668

PR #215, opened 2017-08-22
JBEAP-12515 Redundant config restore before quorum configuration

rhusar:


rhusar:
@nziakova Would this rewording address your concern? Thanks.

nziakova:
This looks good to me, thank you!

PR #226, opened 2017-08-30
JBEAP-12896: Fix typos in messaging output for messaging-clustering-s…

sgilda:
…ingleton, helloworld-mdb, and helloworld-mdb-propertysubstitution quickstarts


sgilda:
@ctomc : There was a merge conflict with that one file. I'm not clear why, because the file looked exactly the same as the eap version before my updates. Anyway, this should be correct, but if you want to redo it, feel free.

ctomc:
I will take a look if maybe we have something out of sync between repos

PR #263, opened 2018-02-27
Jaxrs jwt quickstart

nekdozjam:


jaxrs-jwt/generateKeys.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+echo "Generating private key..."
+ssh-keygen -t rsa -b 2048 -f temp.pem
```
ctomc: This is not acceptable. This is a script that will only work on linux (under some conditions)
we need a java-only solution. java's keytool could probably help here.

jaxrs-jwt/service/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+    limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly.quickstarts</groupId>
+        <artifactId>jaxrs-jwt</artifactId>
+        <version>12.0.0-SNAPSHOT</version>
```
ctomc: we moved to 13, it needs to be rebased.

jaxrs-jwt/service/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+
+        <dependency>
+            <groupId>org.jboss.spec.javax.ws.rs</groupId>
+            <artifactId>jboss-jaxrs-api_2.0_spec</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
+            <version>1.1.2</version>
```
ctomc: never bring such dependency in. we have it manged already under

```
<groupId>org.jboss.spec.javax.json</groupId>
<artifactId>jboss-json-api_1.0_spec</artifactId>
```

Or is there requirement to use 1.1 version from EE 8?

jaxrs-jwt/service/src/main/java/org/jboss/quickstarts/jaxrsjwt/rs/TestRest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @GET
+    @Path("/claims")
+    public Response demonstrateClaims(@HeaderParam("Authorization") String auth) {
+      if (auth != null && auth.startsWith("Bearer ")) {
+            try {
+                JWT j = JWTParser.parse(auth.substring(7));
+              return Response.ok(j.getJWTClaimsSet().getClaims()).build(); //Note: nimbusds converts token expiration time to milliseconds
+            } catch (ParseException e) {
+                e.printStackTrace();
```
ctomc: so even if exception occurs, we return with 200 OK? feels way wrong

jaxrs-jwt/generateKeys.sh
nekdozjam: @ctomc I don't thing that keytool is capable of doing all the work, openssl would still be required to make both keys in correct format. What if I wrote 2 scripts - 1 for windows and 1 for unix, that would use keytool + openssl? Would that be acceptable?

jaxrs-jwt/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+    limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.wildfly.quickstarts</groupId>
+        <artifactId>quickstart-parent</artifactId>
+        <version>13.0.0-SNAPSHOT</version>
```
ctomc: you should use 
  ```
  <parent>
        <groupId>org.wildfly.bom</groupId>
        <artifactId>quickstart</artifactId>
        <version>13.0.0.Alpha3</version>
        <relativePath />
    </parent>
as parent
```
https://github.com/wildfly/quickstart/pull/263/files

PR #290, opened 2018-07-17
Revert "Appy jboss to wfly updates"

emmartins:
Reverts wildfly/quickstart#289


emmartins:
not yet the time for such changes being merged

PR #292, opened 2018-08-13
Fix of README.md wording in jta crash quickstarts

ochaloup:
A small fix in the jta crash quickstarts. There is said the JTA transaction is not distributed over multiple application severs. That's not true - see e.g. https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.1/html/development_guide/java_transaction_api_jta#about_java_transactions_api_jta

```
The TM works in JTA transactions mode, the data is shared in memory,
and the transaction context is transferred by remote EJB calls...
```


emmartins:
@sgilda can you please review?

sgilda:
This looks good to me. Merging.

PR #296, opened 2018-08-20
Documentation clarification and script fix

dvanbalen:
This pull request:

- Documents the need to assign role 'Users' to the quickstart user in order for it to be able to access the servlet via curl
- fixes the module name in remove-custom-module.cli so that it matches that of the added module

### DIFF
--- a/http-custom-mechanism/README.adoc
+++ b/http-custom-mechanism/README.adoc
@@ -67,6 +67,27 @@ include::../shared-doc/back-up-server-standalone-configuration.adoc[leveloffset=
 // Start the {productName} Standalone Server
 include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
 
+[[assign_role_to_the_application_user]]
+== Assign the appropriate Role to the Application User
+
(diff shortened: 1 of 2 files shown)


dvanbalen:
Should've specified the changes are for the http-custom-mechanism quickstart

sgilda:
@dvanbalen : 

You can generate the quickstart README.html files using this command: `mvn clean generate-resources -Pdocs`.

Open the resulting `http-custom-mechanism/README.html` file in a browser and review the section titled `Add the Authorized Application User`. You should see that the command already creates the application user with the `Users` role.

The command to remove the module is also correct. If you look at the command to add the module [1], you will see that it adds the module name `org.jboss.as.quickstart.http_custom_mechanism.custom-http-mechanism`. This matches the name that is removed. 

[1] https://github.com/wildfly/quickstart/blob/master/http-custom-mechanism/configure-elytron.cli#L7

# Add the resource to discover the factory implementation for the custom module.
/subsystem=elytron/service-loader-http-server-mechanism-factory=custom-factory:add(module=org.jboss.as.quickstart.http_custom_mechanism.custom-http-mechanism)

I

sgilda:
@emmartins : Can you close this PR? It is not valid.

PR #299, opened 2018-09-19
WFLY-11041: Update jaxws-retail README to use attribute for product name

sgilda:


sgilda:
@emmartins : This fix also needs to be applied to the JBoss CD 14 release.

PR #300, opened 2018-09-19
JBEAP-15476: Update attributes for CD.14 release

sgilda:


sgilda:
@emmartins : This fix also needs to be applied to the JBoss CD 14 release.

PR #302, opened 2018-09-21
JBEAP-15496: Add requirement to configure authentication to the Red H…

sgilda:
…at Container Registry when deploying to JBoss EAP OpenShift Container Platform


sgilda:
@emmartins : This fix is needed for the CD.14 release also.

PR #303, opened 2018-09-27
JBEAP-15536: Fix the import template URL to use the tagged location f…

sgilda:
…or the CD release


sgilda:
@emmartins : This fix needs to be made to CD14 also.

PR #306, opened 2018-10-11
WFLY-10884: Use the jboss.server.config.dir system property for the f…

sgilda:
https://issues.jboss.org/browse/WFLY-10884


sgilda:
It looks like there was a previous PR to fix the file path issue in the code. Closing this one now.

PR #317, opened 2018-11-02
Update WildFly project and JBoss EAP product versions; update Java version

sgilda:


sgilda:
@emmartins : This update needs to be applied to the JBoss EAP 7.2 quickstarts

PR #319, opened 2018-11-05
JBEAP-15768: Update the root README file to support CD, Product, and Project releases

sgilda:
JIRA: https://issues.jboss.org/browse/JBEAP-15768


sgilda:
@emmartins : 
This fix also needs to be applied to the CD 15 quickstarts. 
It is NOT needed for 7.2.

sgilda:
@emmartins : Hold off on this one for a bit. I need to check on the product naming.

PR #322, opened 2018-11-19
WFLY-11362,JEAP-15877: Add instructions to build quickstart README.html files

sgilda:
JIRA: WFLY-13362
JIRA: JBEAP-15877


sgilda:
@emmartins : This update needs to go into both the wildfly and jboss quickstart repos.

PR #323, opened 2018-11-19
WFLY-11379,JBEAP-15871: Update quickstarts settings.xml with correct repositores

sgilda:
JIRA: https://issues.jboss.org/browse/WFLY-11379

### DIFF
--- a/settings.xml
+++ b/settings.xml
@@ -50,7 +50,7 @@
       <repositories>
         <repository>
           <id>jboss-ga-repository</id>
-          <url>http://maven.repository.redhat.com/techpreview/all</url>
+          <url>https://maven.repository.redhat.com/ga/</url>
           <releases>
             <enabled>true</enabled>
           </releases>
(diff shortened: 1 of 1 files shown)


settings.xml
```diff
@@ -101,12 +101,41 @@
       </pluginRepositories>
     </profile>
 
+    <!-- Configure the JBoss Early Access Maven repository -->
```
nziakova: This is redundant.

settings.xml
```diff
@@ -101,12 +101,41 @@
       </pluginRepositories>
     </profile>
 
+    <!-- Configure the JBoss Tech Preview Maven repository -->
```
nziakova: This is not needed, GA repository should be used instead.

sgilda: @nziakova : The GA repository is listed first. Both the "early access" and "tech preview" repositories are there in case they are ever needed. We have needed them in the past. Should I move the tech preview out of the active profiles section or remove it entirely?

nziakova: techpreview repo has been replaced by the GA one, so it's not needed any more

nziakova: earlyaccess is used for Alpha/Beta releases, ga is used for GA releases

sgilda: I have removed tech preview.

emmartins:
settings.xml was removed

PR #327, opened 2018-12-05
WFLY-11477: Update root README.adoc to be more readable for WildFly and to include build instructions

sgilda:
JIRA: https://issues.jboss.org/browse/WFLY-11477

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,171 +1,524 @@
 include::shared-doc/attributes.adoc[]
 
-= {productNameFull} ({productName}) Quickstarts
+:toc:
+:toclevels: 4
+:numbered:
+
+ifndef::ProductRelease,EAPCDRelease[]
+= WildFly Quickstarts
+
+[toc]
 
 [abstract]
-The quickstarts demonstrate {javaVersion} and a few additional technologies from the {productNameFull} stack. They provide small, specific, working examples that can be used as a reference for your own project.
-
-// The Product builds provide the README.html files. so the following admonition is not needed there.
-ifndef::ProductRelease[]
-[IMPORTANT]
-====
-The quickstart `README` files are written in AsciiDoc, not only because the language is much more powerful than Markdown, but also also because it is possible to extract common instructions into separate files to be reused across the quickstarts. While this makes them more flexible and easier to maintain, unfortunately, included files do not render in a readable format in GitHub or in most text editors.
-
-The Maven plugin that is used to build and deploy the quickstarts can also generate fully rendered `README.html` instructions from the `README.adoc` files. For instructions on how to generate these files, see https://github.com/wildfly/quickstart/blob/master/CONTRIBUTING.adoc#build_the_quickstart_readme_files[Build the README.html Files] in the _CONTRIBUTING_ guide.
-====
-// End ifndef::ProductRelease[]
-endif::[]
-
-//*****************************************************************************
-// Begin Non-CD release Project (WildFly) and Product (JBoss EAP) instructions
-//*****************************************************************************
-ifndef::EAPCDRelease[]
+The quickstarts demonstrate Java EE 8 and a few additional technologies from the WildFly stack. They provide small, specific, working examples that can be used as a reference for your own project.
 
 [[introduction]]
 == Introduction
 
-// JBoss EAP
-ifdef::ProductRelease[]
-These quickstarts run on {productNameFull} {productVersion}. Each quickstart folder contains a `README{outfilesuffix}` file that describes the quickstart features and provides instructions about how to build and run it.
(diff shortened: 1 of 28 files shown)


sgilda:
@emmartins : This PR replaces the one I closed. This version allows us to share the source and you won't have to remember to copy over old source. It only uses the `ProductRelease` and `CDRelease` flags. Both of those flags are commented out for the WildFly release.

Let me know if you have any questions.

README.adoc
```diff
@@ -1,88 +1,323 @@
...
+. Build WildFly core using the following command.
++
+[source,options="nowrap"]
+----
+$ mvn clean install
+----
+
+[discrete]
+[[build_wildfly_server]]
+==== Build the WildFly Server
```
emmartins: When reading I get the impression that I always need to build BOMs, when that is true only when building a Quickstarts development branch (e.g. master) AND it depends on a SNAPSHOT BOMs version ( e.g. https://github.com/wildfly/quickstart/blob/master/pom.xml#L69 ). Maybe rework this content as troubleshooting when build fails?

README.adoc
```diff
@@ -1,88 +1,323 @@
...
+
+[[install_the_quickstart_parent_artifact_in_maven]]
+=== Install the Quickstart Parent Artifact in Maven
+
+A few of the quickstarts require the `quickstart-parent` artifact to be installed in your local Maven repository. To install it, navigate to your __QUICKSTART_HOME__ directory  directory and run the following command.
+
+[source,subs="+quotes,attributes+",options="nowrap"]
+----
+$ cd __QUICKSTART_HOME__
+$ mvn clean install
```
emmartins: should be "mvn clean install -N"

README.adoc
```diff
@@ -1,88 +1,323 @@
...
+
 
 [[run_the_quickstarts]]
-== Run the Quickstarts
+== Run the Quickstarts Using the Maven Command Line
 
 [[build_and_deploy_a_quickstart]]
 === Build and Deploy a Quickstart
 
-The root folder of each individual quickstart contains a `README{outfilesuffix}` file with detailed instructions on how to build and run the example. In most cases you do the following:
```
emmartins: I think the old content here was fine, user should be redirected to specific QS docs.

README.adoc
```diff
@@ -1,88 +1,323 @@
...
+==== Undeploy an Archive
+
+Use the following command to undeploy the quickstart.
+
+[source,options="nowrap"]
+----
+$ mvn wildfly:undeploy
+----
+
+[[verify_the_quickstarts_build_with_one_command]]
```
emmartins: IMHO this kind of content is not for users, content wrt building/deploying/undeploying *all* QS at once  should be in a separated doc, focused on the QS project development.

README.adoc
```diff
@@ -1,88 +1,323 @@
...
+. xref:build_quickstart_readme_files[Build the Quickstart README.html Files]: The quickstart `README` files are written in AsciiDoc to provide modular, reusable content; however, this makes them difficult to read. For this reason, you also must build the quickstart `README.html` files from the AsciiDoc source.
+
+[[build_wildfly_server_and_boms]]
+=== Build the WildFly Server and BOMs
+
+You must build the WildFly server and BOMs to install the Maven artifacts before you can build a quickstart or a quickstart's `README.html` file. The WildFly development team normally takes care of building and publishing the Maven artifacts needed to build the quickstarts in this repository, but as with all new development, there might be times where you need a library that has not yet been updated. Occasionally you might also need an update that was made to WildFly core.
+
+The order for building the WildFly server is as follows.
+
+. xref:build_wildfly_core[Build WildFly core] (if required)
```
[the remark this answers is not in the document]
sgilda: I include this because at one time, I couldn't run a new quickstart without building core. But you are right, it is very rare and it complicates things. I'll remove it.

README.adoc
sgilda: Agree. I have reworked this.

README.adoc
sgilda: The problem is that attributes are not substituted when you read this in GitHub, so I replaced all attributes in the WildFly version of the README.adoc file with actual values.

README.adoc
sgilda: Agree. Removed.

emmartins:
These instructions are not only out of context for users, they are also incorrect, for instance that won’t undeploy everything, it just doesn’t stop when an undeploy fails… Sooner or later we need to clean up this doc and now (start of 7.3 dev) seems a good time!

README.adoc
sgilda: Good catch! Fixed.

sgilda:
Good point. I'll remove the 2 sections on building all quickstarts and undeploying all quickstarts.

PR #330, opened 2018-12-19
Fix Table of Available quickstarts in root README file

sgilda:
@emmartins : This fixes the available table issue. This needs to be applied to the CD15 branch in the EAP quickstarts and there should be no changes required after you merge it there. Prior to this there were 2 available quickstarts table, with attributes determining which table to build. The Maven plugin always just processed the first table and didn't understand the conditionals. I changed it so there is only one table, at the end, for all versions.

I copied this file directly into my checkout of https://github.com/jbossas/eap-quickstarts/tree/CD15, tested it, and it works great.


README.adoc
```diff
@@ -249,113 +249,6 @@ Some of the quickstarts that demonstrate transactions require that you install a
...
-[[available_quickstarts]]
-== Table of Available Quickstarts
-
-All available quickstarts, which are listed in the following table, can be found here: https://github.com/wildfly/quickstart/.
-
-Each quickstart provides the list of technologies demonstrated by the quickstart and the required experience level needed to build and deploy it. Click on the quickstart link in the table to see more detailed information about how to run it. Some quickstarts require deployment of other quickstarts. This information is noted in the `Prerequisites` section of the quickstart `README.html` file.
-
-NOTE: Some of these quickstarts use the H2 database included with WildFly. It is a lightweight, relational example datasource that is used for examples only. It is not robust or scalable, is not supported, and should NOT be used in a production environment!
-
-//<TOC>
```
ctomc: This is a maker that tells build tool to generate table of contents. 
No need to manually update it

sgilda: @ctomc : Yes, I know. But I had 2 tables in the root README and used conditionals for whether it was for the CD release, JBoss EAP, or WildFly. I think that confused the plugin and it wsan't building the correct TOC for the CD release, which only includes a subset of the quickstarts. So I eliminated the confitions and one of the tables and put the other table at the very end. That one is generated for all versions, with no conditions. I n

PR #331, opened 2018-12-20
Update quickstart source GitHub repo URL for CD releases

sgilda:
@emmartins : This fix needs to be applied to the CD15 quickstart repo.


sgilda:
@emmartins : This fix needs to be applied to the CD15 quickstart repo.

PR #341, opened 2019-03-22
[WFLY-11812] command to build docs should be mvn clean package -Pdocs

emmartins:
Issue: https://issues.jboss.org/browse/WFLY-11812

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -151,7 +151,7 @@ To build all of the quickstart `README.html` files, including the root `README.h
 
 [source,options="nowrap"]
 ----
-$ mvn clean generate-resources -Pdocs
+$ mvn clean package -Pdocs
 ----
 
 [TIP]


emmartins:
@sgilda do you foresee any issue with updating the docs build command instructions? The issue is that at some point jboss parent moved to a newer maven-ear-plugin which requires dependency resolution at "generate-resources" goal, so every ear module fails to find its dependencies...

sgilda:
@emmartins : I believe @ctomc came up with that Maven command when he converted the README files from Markdown to AsciiDoc. At that point the old Ruby script that built HTML files no longer worked and Maven was able to handle the build. So I don't see a problem at all. :-)

ctomc:
Generate resouces was used at the time to prevent too much of bulid invoking when building docs.
For example package comes much later in build life cycle so compile, test compile, running tests, assembly of deployments is done as well as building docs.

So moving this to use package phase will result just in much longer and resource intensive build.
Functionally docs will be same.

sgilda:
@ctomc : It's so nice to hear from you! Thanks for the explanation. @emmartins, I gave you an incorrect answer.

emmartins:
retest this please

PR #347, opened 2019-05-07
[WFLY-12053] Update the ejb-security quickstart README to make it more clear that a user needs to be added before running it

fjuma:
https://issues.jboss.org/browse/WFLY-12053

### DIFF
--- a/ejb-security/README.adoc
+++ b/ejb-security/README.adoc
@@ -30,12 +30,16 @@ annotation in the EJB but it should not be granted access to the administrative 
 
 // System Requirements
 include::../shared-doc/system-requirements.adoc[leveloffset=+1]
+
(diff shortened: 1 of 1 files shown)


emmartins:
@fjuma such info seems already "included" by 
`include::../shared-doc/add-application-user.adoc[leveloffset=+1]`
Can you please do `mvn clean package -Pdocs` to build the doc (README.html) and then check its content?

fjuma:
Thanks @emmartins. Verified that the content is correct after building the doc. We had a user run into issues because it wasn't clear from just looking at the README.adoc that a user needed to be added first.

fjuma:
I've updated this PR to just add a blank line between the "includes" to make this stand out more for someone just looking at the .adoc file.

PR #363, opened 2019-09-26
[WFLY-12377] Document usage of JAXWS Client BOM in jaws-* quickstarts

emmartins:
Issue: https://issues.jboss.org/browse/WFLY-12377

### DIFF
--- a/jaxws-addressing/client/pom.xml
+++ b/jaxws-addressing/client/pom.xml
@@ -44,6 +44,12 @@
             <artifactId>jaxws-addressing-service</artifactId>
             <classifier>classes</classifier>
         </dependency>
+        <!-- The JAXWS Client BOM provides application server-compatible dependency management,
+        used for building, testing or debugging JAXWS client applications. Import the BOM to the
+        dependency management section of the app's pom.xml file and then set dependencies on specific artifacts.
+        Alternatively, as shown in the following example, set a single dependency on the BOM for the app to depend
+        on all the artifacts it manages.
+        -->
         <dependency>
             <groupId>org.wildfly</groupId>
             <artifactId>wildfly-jaxws-client-bom</artifactId>
--- a/jaxws-retail/service/pom.xml
+++ b/jaxws-retail/service/pom.xml
@@ -47,7 +47,13 @@
             <artifactId>jboss-ejb-api_3.2_spec</artifactId>
             <scope>provided</scope>
         </dependency>
-        <!-- used to pull correct dependencies for jaxws-tools-maven-plugin  -->
+        <!-- The JAXWS Client BOM provides application server-compatible dependency management,
+        used for building, testing or debugging JAXWS client applications. Import the BOM to the
+        dependency management section of the app's pom.xml file and then set dependencies on specific artifacts.
+        Alternatively, as shown in the following example, set a single dependency on the BOM for the app to depend
+        on all the artifacts it manages.
+        In the following example, the BOM is used to pull the correct dependencies for jaxws-tools-maven-plugin.
+        -->
         <dependency>
             <groupId>org.wildfly</groupId>
             <artifactId>wildfly-jaxws-client-bom</artifactId>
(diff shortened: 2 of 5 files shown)


jaxws-addressing/client/pom.xml
```diff
@@ -44,6 +44,11 @@
             <artifactId>jaxws-addressing-service</artifactId>
             <classifier>classes</classifier>
         </dependency>
+        <!-- The JAXWS Client BOM provides management dependencies compatible with the app server,
+        useful for building, testing or debugging JAXWS client applications. Import it to the app's pom.xml
+        dependencyManagement and then set dependencies on specific artifacts, or alternatively simply set
+        a single dependency on it, for the app to depend on all the artifacts it manages, as done here.
+        -->
```
snelluli1: useful for--> used for
 Import it to the app's pom.xml-->  Import the BOM to the app's pom.xml
dependencyManagement: Is it a section? Can we say "Import the BOM to the dependency management section of the app's pom.xml file"?
The second sentence is too long. I would split it into two: 
"Import the BOM to the dependency management section of the app's pom.xml file and then set dependencies on specific artifacts. Alternatively, as shown in the following example, set a single dependency on the BOM for the app to depend on all the artifacts it manages."
What do you think?

jaxws-retail/service/pom.xml
```diff
@@ -47,7 +47,12 @@
             <artifactId>jboss-ejb-api_3.2_spec</artifactId>
             <scope>provided</scope>
         </dependency>
-        <!-- used to pull correct dependencies for jaxws-tools-maven-plugin  -->
+        <!-- The JAXWS Client BOM provides management dependencies compatible with the app server,
+        useful for building, testing or debugging JAXWS client applications. Import it to the app's pom.xml
+        dependencyManagement and then set dependencies on specific artifacts, or alternatively simply set
+        a single dependency on it for the app to depend on all the artifacts it manages, as done here.
+        In concrete here it's used to pull correct dependencies for jaxws-tools-maven-plugin.
+        -->
```
snelluli1: In concrete here it's used to pull correct dependencies for jaxws-tools-maven-plugin.-->
In the following example, the BOM is used to pull the correct dependencies for jaxws-tools-maven-plugin.

jaxws-addressing/client/pom.xml
snelluli1: @emmartins: The suggestions above are applicable for content in other sections too.

fabiobrz: @emmartins, I agree with the rewording suggested by @snelluli1. 
Besides, I would also change the first sentence from:
"The JAXWS Client BOM provides management dependencies compatible with the app server,"
to 
"The JAXWS Client BOM provides Wildfly application server compatible dependencies management,"
WDYT?

emmartins: @snellulli1 will change all considering your suggestions, thanks
@fabiobrz I would prefer to have no references to the app server "name", since that requires changes of content when applied on EAP builds, so I am thinking on just "The JAXWS Client BOM provides application server compatible dependency management...", is that ok for you?

fabiobrz: @emmartins yes, ok. I agree on stripping the application server name out of the sentence.

fabiobrz: @emmartins, @snelluli1, just a second thought on the above mentioned last sentence. 
Maybe "compliant" would be more appropriate than "compatible"? Just checking here, WDYT?

fabiobrz:
LGTM, I'd wait for a last ack from @snelluli1

emmartins:
I believe compatible is proper in this case, compliant usually implies
there is something like a spec to follow.

fabiobrz:
Right @emmartins, last reply makes sense to me.

jaxws-addressing/client/pom.xml
snelluli1: @fabiobrz: I would say "compatible" is a better word choice. "Compliance" has more to do with legal matters than software systems.

snelluli1: @emmartins, @fabiobrz: Just a minor suggestion. looks good to me otherwise. Thanks!

emmartins:
thanks both

PR #364, opened 2019-10-03
WFLY-12630: Document usage of EJB and JMS Client BOM in Quickstarts

emmartins:
Issue: https://issues.jboss.org/browse/WFLY-12630


emmartins:
@snelluli1 thank you

PR #367, opened 2020-01-22
[WFLY-12737] Add quickstart for MP Health 2.2

xstefank:
https://issues.redhat.com/browse/WFLY-12737


microprofile-health/src/main/java/org/quickstart/microprofile/health/DatabaseConnectionHealthCheck.java
```diff
@@ -0,0 +1,42 @@
...
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+@Readiness
+@ApplicationScoped
+public class DatabaseConnectionHealthCheck implements HealthCheck {
```
jmesnil: I'm not sure that a database connection health check is correct for this quickstart.
To me, it is an anti pattern. Health checks should check the readiness/healthiness of the deployment itself without crossing service boundaries (and the database is another service in that case).

If the application is not considered healthy with that regard, it will be restarted but this will not fix the database connection issue. In production, this would likely cause a cascade of failure/restart from the application without any benefit.

Could the quickstart instead check (and simulate failure) on in-vm resources (thread pools, memory consumption, cpu activity, etc.)?

xstefank: This is a readiness check which in my opinion is directly targeted for this kind of scenario. Readiness check won't restart the container, it just prohibits the traffic. This is the main difference between liveness and readiness. From https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes:

Sometimes, applications are temporarily unable to serve traffic. For example, an application might need to load large data or configuration files during startup, or depend on external services after startup. In such cases, you don’t want to kill the application, but you don’t want to send it requests either.

jmesnil: ah, you're right. I did not pay attention that it was a readiness probe and not a liveness probe.

looks good to me

README.adoc
```diff
@@ -493,6 +493,7 @@ NOTE: Some of these quickstarts use the H2 database included with WildFly. It is
 | link:managed-executor-service/README{outfilesuffix}[managed-executor-service]|EE Concurrency Utilities, JAX-RS, JAX-RS Client API | The `managed-executor-service` quickstart demonstrates how Jakarta EE applications can submit tasks for asynchronous execution. | Beginner | _none_
 | link:messaging-clustering/README{outfilesuffix}[messaging-clustering]|JMS, MDB | The `messaging-clustering` quickstart does not contain any code and instead uses the `helloworld-mdb` quickstart to demonstrate clustering using ActiveMQ Messaging. | Intermediate | link:helloworld-mdb/README.html[helloworld-mdb]
 | link:messaging-clustering-singleton/README{outfilesuffix}[messaging-clustering-singleton]|JMS, MDB, Clustering | The `messaging-clustering-singleton` quickstart uses a JMS topic and a queue to demonstrate clustering using {productName} messaging with MDB singleton configuration where only one node in the cluster will be active. | Advanced | _none_
+| link:microprofile-health/README{outfilesuffix}[microprofile-health]|MicroProfile Health | The `microprofile-health` quickstart demonstates the use of the MicroProfile Health | Beginner | _none_
```
jmesnil: s/demonstates/demonstrates

jmesnil: s/the MicroProfile Health/MicroProfile Health

microprofile-health/README.adoc
```diff
@@ -0,0 +1,425 @@
...
+----
+$ mvn clean package wildfly:deploy
+----
+
+If you now rerun the readiness health check (at `http://localhost:9990/health/ready`)
+the overall `status` should be DOWN. You can also check the liveness check at
+`http://localhost:9990/health/live` which will return the overall `status` UP because
+it isn't influenced by the readiness checks.
+
+As we shouldn't leave this application with a readiness check in a DOWN state and
```
jmesnil: section about quarkus does not apply, right?

xstefank: agh :) thanks a lot!

microprofile-health/README.adoc
```diff
@@ -0,0 +1,424 @@
+include::../shared-doc/attributes.adoc[]
+
+= microprofile-health: MicroProfile Health QuickStart
+:author: Martin Stefanko
+:level: Beginner
+:technologies: MicroProfile Health
+
+[abstract]
+The `microprofile-health` quickstart demonstates the use of the MicroProfile Health
```
darranl: s/demonstates/demonstrates

microprofile-health/README.adoc
```diff
@@ -0,0 +1,424 @@
...
+. Compile the application code and deploy it to {productName}
+[source,options="nowrap"]
+----
+$ mvn clean package wildfly:deploy
+----
+
+// Start the {productName} Standalone Server
+include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
+
+== Running the health check
```
darranl: Should this be === ?

xstefank: can be, thanks

microprofile-health/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+  <name>Quickstart: microprofile-health-quickstart</name>
+  <url>http://maven.apache.org</url>
+
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <management.http.url>http://localhost:9990</management.http.url>
+  </properties>
+
+  <dependencyManagement>
```
darranl: I believe this full dependencyManagement section can go as already defined in the parent.

xstefank: thank you, removed

xstefank:
@emmartins I believe this is ready for review. Only OpenShift tests are missing which are not a part of this PR.

microprofile-health/README.adoc
```diff
@@ -0,0 +1,424 @@
...
+
+Congratulations! You successfully deployed and run this quickstart. If you wish to
+create the application step by step you can follow the subsequent sections.
+
+== Creating the Maven Project
+
+[source,options="nowrap"]
+----
+mvn archetype:generate \
+    -DgroupId=org.wildfly.quickstart.microprofile.health \
```
darranl: Existing quickstarts use a groupID of "org.wildfly.quickstarts" should the projects we generate also use this groupID?

xstefank: Actually, it seems they use `org.jboss.as.quickstarts`. @emmartins should we change packages to `org.jboss.as.quickstarts`?

xstefank: For now, updated to `org.wilfdly.quickstarts`.

emmartins: please use same maven group id as others

xstefank:
@emmartins this is ready

PR #365, opened 2019-11-15
[WFLY-12797/JBEAP-17133] adding elements based on XSD for not having errors in code ready

ochaloup:
https://issues.jboss.org/browse/WFLY-12797

The changes in xml descriptor of `jts` quickstart to avoid errors in Code Ready/Eclipse

PR #369, opened 2020-01-29
[WFLY-13031] EAP quickstart 'messaging-clustering-singleton' shows errors after import

parsharma:
Issue: https://issues.redhat.com/browse/WFLY-13031


emmartins:
btw, I wonder if instead of removing the schema location there is an URL fix, similar to https://github.com/wildfly/quickstart/pull/365 ?

RanabirChakraborty:
> this looks fine to me, yet would like someone else to assure the error is gone on Red Hat Code Ready Studio, @RanabirChakraborty can you do that please since you recently worked on similar issues?

Yeah sure @emmartins I'll help her :)

parsharma:
> btw, I wonder if instead of removing the schema location there is an URL fix, similar to #365 ?

@emmartins  #365 is clone issue. Both issues are same. can we link it?

parsharma:
Hi @emmartins, 
Please let me know do i link both the issues? Because i believed someone is also working on same issue.

PR #370, opened 2020-01-29
WFLY-13035:added <DOCTYPE xml>

sudeshnas93:
Issue: https://issues.redhat.com/browse/WFLY-13035

### DIFF
--- a/ejb-security-context-propagation/src/main/resources/META-INF/jboss-ejb-client.xml
+++ b/ejb-security-context-propagation/src/main/resources/META-INF/jboss-ejb-client.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE xml>
 <!--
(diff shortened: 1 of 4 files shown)


emmartins:
@sudeshnas93 the JIRA refers 4 XML files yet this PR only fixes 2, is that correct?
PS: minor PR/commit format issue, please use "[WFLY-XXXXX] commit msg"

sudeshnas93:
@emmartins Changed rest 2 files as well. Let me know if there needs to be any other change.

emmartins:
@sudeshnas93 now looking at all 4, what's the idea of the DOCTYPE tag, turn off IDE validation?

sudeshnas93:
> @sudeshnas93 now looking at all 4, what's the idea of the DOCTYPE tag, turn off IDE validation?

what I know DOCTYPEs are a key component of compliant web pages. Markup and CSS won’t validate without them but I don't think IDE validation has any relation with it.

PR #371, opened 2020-01-29
WFLY-13021 adding servlet dependency in pom.xml

RanabirChakraborty:
Issue: https://issues.redhat.com/browse/WFLY-13021

### DIFF
--- a/spring-greeter/pom.xml
+++ b/spring-greeter/pom.xml
@@ -66,6 +66,13 @@
             <scope>provided</scope>
         </dependency>
 
+	<!-- Import the javax.servlet-api dependency -->
+        <dependency>
+            <groupId>org.jboss.spec.javax.servlet</groupId>
+            <artifactId>jboss-servlet-api_4.0_spec</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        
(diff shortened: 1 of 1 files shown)


spring-greeter/pom.xml
```diff
@@ -66,6 +66,14 @@
             <scope>provided</scope>
         </dependency>
 
+	<!-- Import the javax.servlet-api dependency -->
+	<dependency>
+		<groupId>javax.servlet</groupId>
+		<artifactId>javax.servlet-api</artifactId>
+		<version>4.0.1</version>
```
darranl: An individual quickstart should not be declaring the version of the dependency, instead for something like the servlet-api the version should be omitted so the version defined in the bom references from the parent project will be used.

emmartins: it should use instead our fork of the servlet specs, i.e

```
        <dependency>
            <groupId>org.jboss.spec.javax.servlet</groupId>
            <artifactId>jboss-servlet-api_4.0_spec</artifactId>
            <scope>provided</scope>
        </dependency>
```

RanabirChakraborty: Added the changes and now it's working fine !

PR #381, opened 2020-02-12
[WFLY-12445] Adds galleon layers tip and parameter needed to enable h…

emmartins:
…2, to CD READMEs
Issue: https://issues.redhat.com/browse/WFLY-12445


emmartins:
Examples of READMEs with the galleon layers tip, and with or without H2 datasource, may be found attached in the linked JIRA

PR #382, opened 2020-02-24
WFLY-12770 Create quickstart for MicroProfile Fault Tolerance 2.1

rhusar:
https://issues.redhat.com/browse/WFLY-12770

Proposal
https://github.com/wildfly/wildfly-proposals/pull/298


microprofile-fault-tolerance/README.adoc
```diff
@@ -0,0 +1,434 @@
...
+        // introduce some artificial failures
+        final Long invocationNumber = counter.getAndIncrement();
+        if (invocationNumber % 4 > 1) { // alternate 2 successful and 2 failing invocations
+            throw new RuntimeException("Service failed.");
+        }
+    }
+}
+----
+
+And inject the code bellow into the `CoffeeResource` endpoint:
```
jmesnil: s/bellow/below

rhusar:
@jmesnil I think that's a good idea to demonstrate how to set this through the MP config subsystem.

microprofile-fault-tolerance/README.adoc
rhusar: Fixed.

microprofile-fault-tolerance/README.adoc
```diff
@@ -0,0 +1,433 @@
...
+
+First, we need a new project. Create a new project with the following command:
+
+// TODO clarify what archetype we can use here
+[source, subs=attributes+]
+----
+mvn archetype:generate \
+    -DarchetypeArtifactId=wildfly-jakartaee-webapp-archetype \
+    -DarchetypeGroupId=org.wildfly.archetype \
+    -DarchetypeVersion=18.0.0.Final
```
pferraro: I don't think you want to use 18.0.0.Final, as this version does not contain fault-tolerance, right?

pferraro: And perhaps parameterize these?

pferraro:
> I think that's a good idea to demonstrate how to set this through the MP config subsystem.

Not via the subsystem per se, but rather via application.properties.

rhusar:
>> I think that's a good idea to demonstrate how to set this through the MP config subsystem.

> Not via the subsystem per se, but rather via application.properties.

The idea proposed by Jeff to demonstrate was how to change things *without* rebuilding or repackaging the application, this I am suggesting how to do it via the subsystem.

microprofile-fault-tolerance/README.adoc
pferraro: Sure, but 19.0.0.Final will be available when this PR is merged.

microprofile-fault-tolerance/README.adoc
```diff
@@ -0,0 +1,433 @@
...
+
+[abstract]
+The `microprofile-fault-tolerance` quickstart demonstrates how to use Eclipse MicroProfile Fault Tolerance in {productName}.
+
+One of the challenges brought by the distributed nature of microservices is that communication with external systems is
+inherently unreliable. This increases demand on resiliency of applications. To simplify making more resilient
+applications, {productName} contains an implementation of the MicroProfile Fault Tolerance specification.
+
+In this guide, we demonstrate usage of MicroProfile Fault Tolerance annotations such as `@Timeout`, `@Fallback`,
+`@Retry` and `@CircuitBreaker`. The specification also introduces `@Bulkhead` and `@Asynchronous` interceptor bindings not
```
tterem: Why are they not covered here?

tterem:
Hi @rhusar, I noticed that other MicroProfile quickstarts contain tests, do you plan to add tests here or is it QE responsibility?

microprofile-fault-tolerance/README.adoc
rhusar: This is whatever Quarkus has, expanding this further I don't consider useful time investment, but contributions are welcome!

rhusar:
Good point, possibly at least a smoke test would be useful – I will add one.

microprofile-fault-tolerance/README.adoc
```diff
@@ -0,0 +1,433 @@
...
+== The Scenario
+
+The application built in this guide simulates a simple backend for a gourmet coffee e-shop. It implements a REST
+endpoint providing information about coffee samples we have on store.
+
+Let's imagine, although it's not implemented as such, that some of the methods in our endpoint require communication
+to external services like a database or an external microservice, which introduces a factor of unreliability.
+
+== Solution
+
```
tterem: It should be mentioned here that fault tolerance subsystem is disabled by default and how to enable it.

(I know that you prepared PR to make it default but I am not sure when it's going to be merged so I mention this in case this quickstart is done sooner)

microprofile-fault-tolerance/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+        </license>
+    </licenses>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.eclipse.microprofile.fault-tolerance</groupId>
+            <artifactId>microprofile-fault-tolerance-api</artifactId>
+            <version>2.1</version>
+            <scope>provided</scope>
+        </dependency>
```
tterem: Adding this dependency should be mentioned in README.

microprofile-fault-tolerance/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-jaxrs</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
tterem: Adding this dependency should be mentioned in README.

microprofile-fault-tolerance/README.adoc
```diff
@@ -0,0 +1,433 @@
...
+in the `CoffeeResource#coffees()` endpoint method in about 50 % of requests.
+
+Why not check that our application works? Deploy the application in a running {productName} server with:
+
+[source,shell]
+----
+mvn package wildfly:deploy
+----
+
+and open `http://localhost:8080/coffee` in your browser. Make couple of requests (remember, some of them we expect
```
tterem: `http://localhost:8080/coffee`-> `http://localhost:8080/microprofile-fault-tolerance/service/coffee`

Same for all other similar links in this document.

tterem:
Hi @rhusar, I've added few comments about what needs to be added to README.
Also, step with Circuit Breaker is not working for me but for now I am not sure why.

microprofile-fault-tolerance/src/main/java/org/wildfly/quickstarts/microprofile/faulttolerance/CoffeeResource.java
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * Returns how many packages of given coffee we have on store.
+     *
+     * In this case, the failures are introduced inside the CDI bean this method calls. The bean will repeat two
+     * successful and two failed invocations. It also defines a {@link CircuitBreaker} policy, which will open a circuit
+     * breaker after two failed invocations, and then stay open for 5 seconds.
+     */
+    @Path("/{id}/availability")
+    @GET
+    public Response availability(@PathParam int id) {
```
tterem: `@PathParam` -> `@PathParam("id")`

This fixes Circuit Breaker issue that I mentioned. 

Now, the id is always 0 and there is no coffee with id == 0. They have ids 1, 2 and 3. In other usages of PathParam, id is 0 as well but it doesn't break functionality of the quickstart entirely. For example recommendations(@PathParam int id) method calls coffeeRepository.getRecommendations(id) which only checks coffees by non-equality of the id so it always picks first two coffees to show. Correctly, it should show coffees 1 and 3 if called with id = 2.

That being said, this code is technically correct, but it requires some additional configuration, as mentioned here:

https://docs.jboss.org/resteasy/docs/4.5.2.Final/userguide/html_single/index.html#_NewParam

microprofile-fault-tolerance/README.adoc
rhusar: Are we sure? There aren't any for WildFly 9, 10, 11, 12, 13, 14, 15, 16, nor 17.

microprofile-fault-tolerance/README.adoc
rhusar: PS: contributions are welcome. These documents are not immutable ;-)

microprofile-fault-tolerance/README.adoc
pferraro: Oh for f@#$ sake...

rhusar: Just an update from the call: we are going to use a generic archetype and add a section how to import a bom.

rhusar: Done.

microprofile-fault-tolerance/src/main/java/org/wildfly/quickstarts/microprofile/faulttolerance/CoffeeApplication.java
```diff
@@ -0,0 +1,8 @@
+package org.wildfly.quickstarts.microprofile.faulttolerance;
+
+import javax.ws.rs.ApplicationPath;
+import javax.ws.rs.core.Application;
+
+@ApplicationPath("/service")
+public class CoffeeApplication extends Application {
```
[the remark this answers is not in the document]
rhusar: Done.

microprofile-fault-tolerance/README.adoc
rhusar: Done. Added steps.

microprofile-fault-tolerance/pom.xml
rhusar: Done. Added a section to add to pom.

microprofile-fault-tolerance/pom.xml
rhusar: Done.

microprofile-fault-tolerance/README.adoc
rhusar: Done. I have actually removed the extra /service from the URL.

microprofile-fault-tolerance/src/main/java/org/wildfly/quickstarts/microprofile/faulttolerance/CoffeeResource.java
rhusar: Good point, talked to RESTeasy team, we don't need to use the fancy autogenerated stuff. Done now.

rhusar:
Cleared stale review status.

rhusar:
retest this please

PR #385, opened 2020-02-27
Next is 19.0.0.Beta3

Belak-Ghouas:
Hello , 
Problem in pom.xml , with dependencies.


emmartins:
At development branches you need to build first the same name branches of https://github.com/wildfly/wildfly and https://github.com/wildfly/boms

PR #368, opened 2020-01-28
[WFLY-12901] Add a new quickstart covering MicroProfile JWT.

darranl:
Includes a complete example and instructions to create the example from scratch.
Also contains a test case to enable testing the interaction with the example.

https://issues.redhat.com/browse/WFLY-12901

Analysis - https://github.com/wildfly/wildfly-proposals/pull/273

Depends on https://github.com/wildfly/wildfly-archetypes/pull/6

PR #392, opened 2020-03-17
Fix module name template in template's pom.xml

rhusar:


rhusar:
As pointed out by Darran when discussing with him on here https://github.com/wildfly/quickstart/pull/368#discussion_r393679129

rhusar:
@emmartins do you want a backport to 19?

PR #399, opened 2020-03-18
[19.0.x] [WFLY-13258] replace commons-logging

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-13258

### DIFF
--- a/microprofile-config/pom.xml
+++ b/microprofile-config/pom.xml
@@ -61,9 +61,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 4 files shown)


emmartins:
@xstefank please review

jamezp:
It doesn't really matter, but these seem to be used for tests so it should be fine to use the default commons-logging. We only replace this in WildFly as it requires an implementation of where logs should be written.

emmartins:
@jamezp at least we get ridden of an unmanaged dependency version

PR #400, opened 2020-03-19
[WFLY-13258] replace commons-logging

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-13258


jamezp:
I think we need to be careful replacing dependencies like this. It's not necessarily meant for end users as it requires jboss-logging as well.

PR #401, opened 2020-03-23
WFLY-13268 MP Fault Tolerance quickstart cannot resolve dependencies …

rhusar:
…with wildlfy-microprofile BOM

Jira
https://issues.redhat.com/browse/WFLY-13268

Downstream
https://github.com/wildfly/quickstart/pull/402

### DIFF
--- a/microprofile-fault-tolerance/pom.xml
+++ b/microprofile-fault-tolerance/pom.xml
@@ -52,6 +52,15 @@
 
     <dependencyManagement>
         <dependencies>
+            <!-- importing the jakartaee8-with-tools BOM adds specs and other useful artifacts as managed dependencies -->
+            <dependency>
+                <groupId>org.wildfly.bom</groupId>
+                <artifactId>wildfly-jakartaee8-with-tools</artifactId>
+                <version>${version.server.bom}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <!-- importing the microprofile BOM adds MicroProfile specs -->
             <dependency>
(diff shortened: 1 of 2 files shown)


microprofile-fault-tolerance/pom.xml
```diff
@@ -52,6 +52,13 @@
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
```
xstefank: you don't need the whole dependencyManagement section here as both boms are in parent.

rhusar: That's done towards the efforts of removing the boms from parent as they should not be there.

rhusar:
+1 @xstefank I agree that the `widlfly-microprofile` BOM should be sufficient and we should not mix and match.

microprofile-fault-tolerance/pom.xml
rhusar: Resolving discussion - cc @emmartins

emmartins:
IMHO You should explain why you importing the BOMs, e.g. XML comment before each import (in both the pom.xml and README.adoc XML snippet)

rhusar:
@emmartins I could but this doesn't seem to be a standard, e.g. https://github.com/wildfly/quickstart/blame/master/microprofile-jwt/README.adoc#L180-L200

emmartins:
@rhusar it's something that should be done for all quickstarts once BOMs imports are added

rhusar:
@emmartins OK, added a note.

PR #404, opened 2020-03-23
[19.x] Upgrade wildfly-maven-plugin and fix WildFly version

jamezp:
JIRA's:

- https://issues.redhat.com/browse/WFLY-13271
- https://issues.redhat.com/browse/WFLY-13272

Upstream #403

### DIFF
--- a/deltaspike-authorization/pom.xml
+++ b/deltaspike-authorization/pom.xml
@@ -191,7 +191,6 @@
             <plugin>
                 <groupId>org.wildfly.plugins</groupId>
                 <artifactId>wildfly-maven-plugin</artifactId>
-                <version>${version.wildfly.maven.plugin}</version>
             </plugin>
         </plugins>
     </build>
--- a/http-custom-mechanism/pom.xml
+++ b/http-custom-mechanism/pom.xml
@@ -75,7 +75,6 @@
             <plugin>
                 <groupId>org.wildfly.plugins</groupId>
                 <artifactId>wildfly-maven-plugin</artifactId>
-                <version>${version.wildfly.maven.plugin}</version>
             </plugin>
         </plugins>
     </build>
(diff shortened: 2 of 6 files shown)


emmartins:
@jamezp do we ever expect different versions between BOM and server? If not then perhaps rename property version.server.bom to version.server and use it on both the BOM import, and in the wildfly maven plugin

jamezp:
@emmartins Given the above issue I would say yes :) The `version.wildfly.bom` is set to `19.0.1.Final` in the tag and there is not WildFly 19.0.1.Final.

emmartins:
@jamezp ok, and what about EAP branches, what would be the config for the plugin?

jamezp:
@emmartins I'm not really sure I follow. Do you mean the change to the fault tolerance plugin configuration?

emmartins:
@jamezp I mean will the plugin be able to be configured for EAP or EAP MP Quickstarts will have to use WildFly 19.0.0.Final as configured here?

emmartins:
also doesn't sound great that 19.1.0.Final Quickstarts will use WFLY 19.0.0.Final, won't the plugin be able to use 19.1.0.Final once released?

jamezp:
The plugin doesn't care what version of the server is used. It can be WildFly whatever or EAP whatever. The version property here of 19.0.0.Final is used in fault tolerance if the user uses `mvn wildfly:run`. Otherwise the version isn't used. Really now the version could be left off completely. If the `mvn wildfly:run` command is ran it will determine the newest version of WildFly to used based on JBoss Stacks.

emmartins:
@jamezp considering QS is released after WildFly, can't we assume the released WildFly is available to be used in the plugin and always use that?
@rhusar @pferraro can you please assert if the wildfly version is really needed for the fault tolerance and openapi quickstarts work correctly?

jamezp:
We assume it's released, we can't assume JBoss Stacks has been updated to download the latest version.

rhusar:
@emmartins IIRC than yes, because the version of that wildfly maven plugin would download an older version which does not have these subsystems...

emmartins:
@rhusar ok thanks
@jamezp please just rename the wildfly.version property to version.server then

emmartins:
@jamezp same in related master PR please

PR #413, opened 2020-05-12
WFLY-13458 Issues in readme for microprofile-health QS

xstefank:
https://issues.redhat.com/browse/WFLY-13458

### DIFF
--- a/microprofile-health/README.adoc
+++ b/microprofile-health/README.adoc
@@ -25,22 +25,23 @@ In this quickstart, we have a simple REST application that exposes MicroProfile 
 functionalities at the `/health/live` and `/health/ready` endpoints according to the
 specification.
 
-== Solution
+// System Requirements
+include::../shared-doc/system-requirements.adoc[leveloffset=+1]
 
-We recommend that you follow the instructions in the next sections and create the
-application step by step. However, you can go right to the completed example which
-is available in this directory.
-
-. Start your {productName} server
-
-. Compile the application code and deploy it to {productName}
-[source,options="nowrap"]
-----
-$ mvn clean package wildfly:deploy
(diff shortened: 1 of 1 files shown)


xstefank:
@nziakova can you take a look, please?

nziakova:
@xstefank Thank you for the fixes. It looks better now, although I can only tell from which shared resources have been added. What I am still missing is (3) from JIRA - "The checks array is empty as we have not specified any health check procedure yet so let's define some." is not correct because it comes after the complete example deployment - please correct me if I am wrong.

xstefank:
@nziakova I've included Undeploy section before the Creating project from scratch section. So the user following section by section should undeploy the full quickstart before proceeding with create from scratch. That should solve the issue.

However, in the Solution section I want to point people to either full quickstart (only deploy) **OR** Create project from scratch. It didn't occur to me that someone will do both :) Can this be communicated better?

xstefank:
And you can generate the README.html with `mvn clean package -Pdocs` if you would like to check the full expanded sections.

nziakova:
I think it is pointed out fine that user should follow complete example OR step-by-step one. Still the sentence about empty checks is between deploying and undeploying the complete example, so if someone follows the complete example instructions, they will get different output than documented.

xstefank:
@nziakova thank you. I see what I missed now. I just removed the sentence as it's not that important.

xstefank:
@emmartins @nziakova I believe everything is fixed now and this can be merged.

emmartins:
looks good to me, if @nziakova has nothing against I will merge it shortly

nziakova:
Nothing from me, feel free to merge.

PR #414, opened 2020-05-13
WFLY-13462 Issues in readme for microprofile-config QS

xstefank:
https://issues.redhat.com/browse/WFLY-13462

### DIFF
--- a/microprofile-config/README.adoc
+++ b/microprofile-config/README.adoc
@@ -11,6 +11,7 @@ The `microprofile-config` quickstart demonstrates the use of the MicroProfile Co
 :standalone-server-type: default
 :archiveType: jar
 :archiveName: {artifactId}
+:arq-prerequisities: These tests require that JBOSS_HOME environment variable is set.
 
 == What is it?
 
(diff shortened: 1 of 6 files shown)


xstefank:
@nziakova can you also review, please?

nziakova:
@xstefank Thank you. It looks good, just maybe consider changing the order of undeploy and run arq tests sections as I noted in https://issues.redhat.com/browse/JBEAP-19461 I'd put  the JBDS section after these all, something like here https://github.com/wildfly/quickstart/tree/master/kitchensink - but that's just a suggestion, it will work without it as well.

xstefank:
@nziakova updated in both config and health. Thank you.

xstefank:
@emmartins we can't do attribute expansion in the code blocks so I changed everything to code block from `in WildFly` to `in the server`. If you can do the proper expansion in the pom.xml I think we should keep it at least there. Otherwise, this PR should be good to go.

microprofile-health/README.adoc
```diff
@@ -10,7 +10,8 @@ The `microprofile-health` quickstart demonstrates the use of the MicroProfile He
 
 :standalone-server-type: default
 :archiveType: jar
-:archiveName: {artifactId}-service
+:archiveName: {artifactId}
+:jbossHomeName: JBOSS_HOME
```
emmartins, line 14: not needed, right?

xstefank: ah, I forgot, thanks

microprofile-config/README.adoc
```diff
@@ -11,6 +11,7 @@ The `microprofile-config` quickstart demonstrates the use of the MicroProfile Co
 :standalone-server-type: default
 :archiveType: jar
 :archiveName: {artifactId}
+:arq-prerequisities: These tests require that {jbossHomeName} environment variable is set.
```
emmartins, line 14: Why not `JBOSS_HOME` instead of {jbossHomeName}, as used in run-arquillian-tests.adoc bottom note? The way I see jbossHomeName is just a prod vs community differentiation for the app server distribution directory, and should not be confused with the env property JBOSS_HOME.

xstefank: The tests require JBOSS_HOME, it can't be EAP_HOME which would be in jbossHomeName. I can change it to JBOSS_HOME env var OR jbossHome sys prop but that seems too much information and not much value. Should I change it?

emmartins: I mean this line to be instead:
:arq-prerequisities: These tests require that `JBOSS_HOME` environment variable is set.

xstefank: thanks, I understand now

emmartins:
@nziakova do you want to try this as it is now, or should I proceed with merge and cherry-pick to XP?

nziakova:
Feel free to merge, my note was really minor. Will try full instructions in candidate release.

PR #422, opened 2020-05-15
WFLY-13473 Missing repositories in generated poms for MP QS in EAP

xstefank:
https://issues.redhat.com/browse/WFLY-13473

### DIFF
--- a/microprofile-config/README.adoc
+++ b/microprofile-config/README.adoc
@@ -129,6 +129,9 @@ the generated build section):
 </build>
 ----
 
+// Setup required repositories
+include::../shared-doc/setup-repositories.adoc[leveloffset=+1]
+
 As this is a Jakarta REST application we need to also create an application class.
 Create `org.wildfly.quickstarts.microprofile.config.JaxRsApplication` with the following content:
 
--- a/microprofile-fault-tolerance/README.adoc
+++ b/microprofile-fault-tolerance/README.adoc
@@ -141,6 +141,9 @@ Add the following section under configuration:
 </build>
 ----
 
+// Setup required repositories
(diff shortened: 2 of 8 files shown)


rhusar:
@xstefank Needs a rebase, probably just a clean one but seems to have triggered a merge conflict.

xstefank:
Fixed the conflict.

emmartins:
@nziakova  can you please review?

nziakova:
The change looks good, thank you. One note to possibly consider - it is a rather large amount of text copied to 7 (so far) readmes. Would it make sense to put it to shared-resources and just include it in these readmes? This way if similar QS is created in the future (with from-scratch instructions) the author could simply include prepared text without the risk of making copy-paste errors. Also could be better from maintenance point of view.

nziakova:
Another option would be not including the prepared text directly in readme, just linking a doc which shows how to do that - could be better from readability point of view not to make the readme too long. I don't have strong opinion on which one would be better though.

emmartins:
@nziakova @xstefank I believe a link jump would make it harder to follow, I agree using a shared-docs section  is an improvement tho.

xstefank:
done in the added file (option 1).

nziakova:
@emmartins I think you are right and having to follow a link would really make it harder to follow. Just wanted to show another option because the instructions are getting long. However, it's a step-by-step guide so it should not be an issue for users that chose that option (there's still the short path with complete example).

emmartins:
@darranl @rhusar @pferraro since this will be applied to your QS too, anything against it?

rhusar:
@emmartins Looks good assuming Martin knows what he is doing :-)

PR #429, opened 2020-05-20
WFLY-13496 Change the packaging of MP Health quickstart to ejb to ena…

xstefank:
…ble deployment through IDEs

https://issues.redhat.com/browse/WFLY-13496

### DIFF
--- a/microprofile-health/README.adoc
+++ b/microprofile-health/README.adoc
@@ -97,7 +97,18 @@ Open the project in your favourite IDE.
 
 Open the generated `pom.xml`.
 
-The first thing to do is to setup our dependencies. Add the following section to your
+The first thing to do is to change the packaging to `war` as this is required by 
+the IDEs to recognize the application as a deployment:
+
+[source,xml]
+----
+<packaging>war</packaging>
+----
+
+NOTE: For non-IDE deployments, the plain `jar` packaging is sufficient for
+the MicroProfile Health applications.
+
+Next we need to setup our dependencies. Add the following section to your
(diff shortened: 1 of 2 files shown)


microprofile-health/pom.xml
```diff
@@ -13,7 +13,7 @@
   </parent>
 
   <artifactId>microprofile-health</artifactId>
-  <packaging>jar</packaging>
+  <packaging>ejb</packaging>
```
emmartins, line 16: with this documentation change shouldn't the pom packaging stay as jar?

emmartins: ok, think I understand now, the generated pom.xml has no such change, right?

xstefank: There is no way to generate the packaging directly as ejb, so users for step-by-step instructions need to do the change themselves.

This change is required as the QS itself is consumed in the JB studio. They don't follow step-by-step AFAIK.

microprofile-health/README.adoc
```diff
@@ -97,7 +97,18 @@ Open the project in your favourite IDE.
 
 Open the generated `pom.xml`.
 
-The first thing to do is to setup our dependencies. Add the following section to your
+The first thing to do is to change the packaging to `ejb` as this is required by
```
emmartins, line 100: please change `ejb` to `war`

microprofile-health/pom.xml
emmartins: please change to war

PR #434, opened 2020-06-10
[WFLY-13582] Mention sampler configuration in MP Opentracing Quicksta…

jstourac:
…rt readme.

Issue: https://issues.redhat.com/browse/WFLY-13582

### DIFF
--- a/microprofile-opentracing/README.adoc
+++ b/microprofile-opentracing/README.adoc
@@ -36,6 +36,12 @@ include::../shared-doc/use-of-jboss-home-name.adoc[leveloffset=+1]
 // Start the {productName} Standalone Server
 include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
 
+Please note the additional parameters defined for the server startup script.
+These alternate some defaults of Jaeger service in {productName} server which are
+configured for production where the traffic is usually much bigger. As described
+on https://www.jaegertracing.io/docs/1.16/troubleshooting/[Jaeger troubleshooting],
+for development we shall alternate default sampling strategy and span logging at least.
+
 == Solution
 
 We recommend that you follow the instructions that
@@ -179,7 +185,6 @@ Now you can access `http://localhost:16686` in your browser to see the Jaeger UI
 console.
 
 Now we can start adding our custom spans from our application.
-
 
 == Implicit tracing of REST resources
 


emmartins:
@xstefank can you pls look at this? One issue I notice is use of WildFly instead of {productName}, and $WILDFLY_HOME should be parameterized too, but perhaps no need for a new doc section, just a tip somewhere?

emmartins:
@xstefank what about turning this into an asciidoc tip instead of a new section?

xstefank:
I originally mentioned that only in the Running the server section, however, this is indeed important as without it traces aren't collected. So I think that the section will better highlight this fact.

jstourac:
Hi @emmartins and @xstefank, thanks for the review! I've updated this MR just now based on your comments replacing WildFly and WILDFLY_HOME by a parameters. Feel free to review again, thanks!

emmartins:
@xstefank think at least we need to use a different title, this will confuse readers since it instructions to run the server are elsewhere.

jstourac:
@emmartins, any proposals? Does the `Configuration of Jaeger service on server side` title sound good to you?

emmartins:
@jstourac so I just built the README.adoc of your branch (try "mvn clean generate-resources -Pdocs" in MicroProfile-opentracing dir and then open the README.html), and I see that such params are actually already in the server startup command. After a chat with @xstefank we agree it is better to just add an asciidoc "NOTE:" after the include of the server startup section (line 37).

In concrete I would like to ask you to use "NOTE:" instead of a new section, place it under the include line, and of course rework a bit the text since the command with the parameters is already there.

jstourac:
@emmartins, thanks for the notice - truth is that I haven't seen the HTML version of readme till now and I really didn't know anything like this exist :) Now I understand your objections. I've updated this as you suggested (hopefully), although the note in HTML file is weirdly aligned. I believe this is because the first note is part of the included file. Not sure how to align it properly, though. To be honest, with the latest info, I can be persuaded that this MR is kind of redundant now. Let me know what you think :)

microprofile-opentracing/README.adoc
```diff
@@ -36,6 +36,12 @@ include::../shared-doc/use-of-jboss-home-name.adoc[leveloffset=+1]
 // Start the {productName} Standalone Server
 include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
 
+NOTE: Please note the additional parameters defined for the server startup script.
```
emmartins, line 39: just drop the "NOTE:", agree the alignment looks weird, your text already has "Please note" anyway :-)

jstourac: @emmartins , oh, you are right :) I've just updated, looks much better now. Ready for re-review :)

emmartins:
thanks @jstourac

PR #444, opened 2020-09-17
[WFLY-11492] Quickstart http-custom-mechanism: documentation step fails

SoniaZaldana:
https://issues.redhat.com/browse/WFLY-11492

### DIFF
--- a/http-custom-mechanism/configure-security-domain.cli
+++ b/http-custom-mechanism/configure-security-domain.cli
@@ -2,8 +2,9 @@
 
 # Start batching commands
 batch
-#  Map the application security domain to an Elytron HTTP authentication factory.
-/subsystem=undertow/application-security-domain=other:add(http-authentication-factory=application-http-authentication)
+
+#  Map the application security domain to an Elytron Security Domain.
+/subsystem=undertow/application-security-domain=other:add(security-domain=ApplicationDomain)
 
 # Run the batch command
 run-batch
--- a/http-custom-mechanism/README.adoc
+++ b/http-custom-mechanism/README.adoc
@@ -69,7 +69,7 @@ include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
 [[configure_the_application_security_domain]]
 == Configure the Application Security Domain
 
-By default, the deployed application uses the `other` security domain. You need to map this security domain to an Elytron HTTP authentication factory. For your convenience, this quickstart includes the command to configure the security domain in the `configure-security-domain.cli` script located in the root directory of this quickstart.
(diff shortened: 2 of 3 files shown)


emmartins:
@darranl can you pls review?

http-custom-mechanism/configure-security-domain.cli
```diff
@@ -2,7 +2,11 @@
 
 # Start batching commands
 batch
-#  Map the application security domain to an Elytron HTTP authentication factory.
+
+# Configure an http authentication factory
+/subsystem=elytron/http-authentication-factory=application-http-authentication:add(http-server-mechanism-factory=global,security-domain=ApplicationDomain,mechanism-configurations=[{mechanism-name=BASIC,mechanism-realm-configurations=[{realm-name=Simple Realm}]},{mechanism-name=FORM}])
```
fjuma, line 7: @SoniaZaldana The default WildFly config used to contain an `application-http-authentication` HTTP authentication factory. A little while ago, we removed this from the default config and we added the ability to specify a `security-domain` directly from the `application-security-domain` resource in the Undertow subsystem instead of needing to reference an `http-authentication-factory`. For simple cases where a deployment can make use of the standard HTTP authentication mechanisms (i.e., BASIC, CLIENT_CERT, DIGEST, FORM), we can map the application security domain to an Elytron security domain as follows:

```
/subsystem=undertow/application-security-domain=other:add(security-domain=ApplicationDomain)
```

Since the first part of the quickstart demonstrates how to use BASIC authentication, we should use this simple command instead of adding the `http-authentication-factory` and referencing that from the `application-security-domain` mapping.

The second part of the quickstart then shows how to make use of a custom HTTP authentication mechanism. In this case, `configure-elytron.cli` shows how to add an appropriate `http-authentication-factory` and update the `application-security-domain` mapping to reference that. This script will need to be updated to undefine the `security-domain` attribute from the `application-security-domain` mapping prior to configuring the `http-authentication-factory` attribute.

fjuma (changes requested):

http-custom-mechanism/configure-security-domain.cli
SoniaZaldana: Thanks for the through explanation! I've updated the code.

fjuma: Thanks, Sonia! Looks good. Just one more minor change will be needed to update the following sentence in the "Configure the Application Security Domain" section in the README:

"You need to map this security domain to an Elytron HTTP authentication factory."

SoniaZaldana: Oh right, sorry I missed that. I updated the README.

fjuma (approved):

PR #466, opened 2021-02-05
microprofile-config: Add <cloud /> config to the bootable-jar profile

jmesnil:


jmesnil:
@jfdenise should we apply that to all quickstarts that have a `bootable-jar` profile?

jfdenise:
@jmesnil , we already have a profile that targets cloud, bootable-jar-openshift. If we want a profile that uses <cloud> but not JKube, we need a new profile something like: `bootable-jar-cloud`

jmesnil:
could we split the existing bootable-jar-openshift in two?
* `bootable-jar-openshift` that is equivalent to `bootable-jar` + `<cloud />`
* `bootable-jar-jkube` that triggers https://github.com/wildfly/quickstart/blob/4d3461d5e10798eb10715ce53811815f3aa06885/microprofile-config/pom.xml#L142-L164

This way, an user wanted to build the bootable-jar for openshift would have to run `mvn package -P bootable-jar-openshift -P bootable-jar-jkube`.
Someone (like me) that wants to deploy the bootable-jar on OpenShift with a build config would run `mvn package -P bootable-jar-openshift`.

wdyt?

jfdenise:
That would work, the concern I have is that we would change the behavior of the existing bootable-jar-openshift profile. quickstart doc would need to be updated. @emmartins what do you think of such incompatible change?

emmartins:
are those alternatives, i.e. either use <cloud/> or jkube? If that's the case couldn't we continue have a single bootable-jar-openshift, and in case <cloud/> doesn't work for all then it uses jkube instead for those, and the asciidoc related section could have conditional content for each alternative

microprofile-config/pom.xml
```diff
@@ -104,6 +104,9 @@
                               <layer>microprofile-platform</layer>
                           </layers>
                           <cloud/>
+                          <plugin-options>
+                            <jboss-fork-embedded>true</jboss-fork-embedded>
```
jfdenise, line 108: I think that we should have a property there.

jfdenise:
@jmesnil , using -Djkube.skip=true fixed your concern. This PR can be closed.

emmartins:
@jmesnil @jfdenise I am bit confused on this one, do we need it or not?

jmesnil:
this is no longer need as JKube can be disabled.

PR #467, opened 2021-02-05
microprofile-config: add homepage

jmesnil:
Display a list of the available endpoints when the application is
opened.


emmartins:
@jmesnil @ehsavoie IMHO this could also rework the README.adoc , and simplify both the instructions there and the user experience

PR #469, opened 2021-02-18
[WFLY-14398] Update microprofile-health arquillian tests to use stan…

xstefank:
…dalone-microprofile.xml config

https://issues.redhat.com/browse/WFLY-14398


xstefank:
This should go into MP4 branch.

PR #477, opened 2021-03-12
[WFLY-14563] todo-backend quickstart

jmesnil:
Add a new quickstart `todo-backend` to showcases a typical use of
WildFly: a backend that exposes a Web API (using JAX-RS) and connects to
a DB (using JPA).

The quickstart focuses on local and cloud deployment with Bootable Jar
and showcase the following features:
* use the wildfly-datasources-galleon-pack to connect to an external
  PostgreSQL database
* Execute a CLI script at build time to specialize the WildFly runtime
  provisioned by the Bootable Jar
* Use of Helm Charts for cloud deployment

JIRA: https://issues.redhat.com/browse/WFLY-14563


jmesnil:
Dev analysis is at  https://github.com/wildfly/wildfly-proposals/pull/380

pom.xml
```diff
@@ -80,6 +80,9 @@
         <version.server.bom>24.0.0.Beta1-SNAPSHOT</version.server.bom>
         <version.microprofile.bom>24.0.0.Beta1-SNAPSHOT</version.microprofile.bom>
 
+        <version.server.bootable-jar>23.0.0.Final</version.server.bootable-jar>
```
jbliznak, line 83: Any reason for value different from the properties just above?

todo-backend/README.adoc
```diff
@@ -0,0 +1,302 @@
...
+by default:
+
+[source,options="nowrap"]
+----
+$ mvn clean package
+----
+
+== Run the backend locally
+
+=== Local PostGreSQL database
```
jbliznak, line 122: typo capital `G`

todo-backend/README.adoc
```diff
@@ -0,0 +1,302 @@
...
+[source,options="nowrap"]
+----
+$ helm repo add wildfly https://docs.wildfly.org/wildfly-charts/
+"wildfly" has been added to your repositories
+$ helm search repo wildfly
+NAME                    CHART VERSION   APP VERSION     DESCRIPTION
+wildfly/wildfly         1.1.0           22.0            Build and Deploy WildFly applications on OpenShift
+----
+
+=== Deploy a PostGreSQL Database on OpenShift
```
jbliznak, line 191: typo capital `G`

todo-backend/todo-backend.yaml
```diff
@@ -0,0 +1,43 @@
+# This configuration file can be used to build and deploy the todo-backend
+# quickstart on OpenShift with the Helm Chart for WildFly.
+build:
+  uri: https://github.com/jmesnil/quickstart.git
```
jbliznak, line 4: this must be changed before merging

todo-backend/todo-backend.yaml
```diff
@@ -0,0 +1,43 @@
...
+  ref: todo-backend
+  mode: bootable-jar
+  env:
+    - name: ARTIFACT_DIR
+      value: todo-backend/target
+    - name: MAVEN_ARGS_APPEND
+      value: -am -pl todo-backend -P bootable-jar-openshift -Dversion.server.bom=23.0.0.Final -Dversion.microprofile.bom=23.0.0.Final
```
jbliznak, line 11: We need to get rid of the versions or somehow parameterize them

jbliznak, line 11: Another comment:
> The backend is packaged as a Bootable Jar and configured to be deployable on OpenShift with the Maven Profile `bootable-jar-openshift` which is active
by default:

so there should be no need to specify it here (unless OpenShift adds something more which would disable the `active by default` feature

jbliznak (changes requested):

pom.xml
jmesnil: yes: the quickstart relies on snapshot of WildFly server but on release of the Bootable Jar plug-in.

(e.g. https://github.com/wildfly/quickstart/blob/52fd07d85208df512c7f0a31d20f3fd614f9d58f/microprofile-config/pom.xml#L23)

I'm moving this property to the root `pom.xml` to avoid having to repeat that in all quickstarts that use bootable jar.

todo-backend/README.adoc
jmesnil: fixed, thanks

todo-backend/todo-backend.yaml
jmesnil: fixed, thanks

todo-backend/todo-backend.yaml
jmesnil: well... this is silly.

We can not build the image from the quickstart master branch as it relies on SNAPSHOTS of WildFly.
So if I want to build the image, I must pass Final release for WildFly and BOMS.

Now, once we have a tag with the quickstart (that uses Final release of WildFly), we will not need to override them... but we will have to update this YAML for every new release of the quickstart...

jmesnil: fixed, thanks

jmesnil:
@jbliznak I replaced most occurrences of `WildFly` in the README with the `{productName}` macro. The remaining occurrences will be replaced manually when we add this quickstart to EAP quickstarts.

todo-backend/todo-backend.yaml
```diff
@@ -0,0 +1,43 @@
...
+# quickstart on OpenShift with the Helm Chart for WildFly.
+build:
+  uri: https://github.com/quickstart/quickstart.git
+  ref: todo-backend
+  mode: bootable-jar
+  env:
+    - name: ARTIFACT_DIR
+      value: todo-backend/target
+    - name: MAVEN_ARGS_APPEND
+      value: -am -pl todo-backend
```
jbliznak, line 11: One more question to this, before we were using `GIT_CONTEXT_DIR` in template here for this so we are changing this by design?

Because there is a difference between these too procedures:
```
cd todo-backend
mvn install ...
```

```
mvn install -am -pl todo-backend
```
* In the former, if `todo-backend` contains any submodules, these would be build correctly.
* For the latter, if `todo-backend` contains any submodules, these would not be build, only the top `todo-backend` would
  Also in this case, the project root pom.xml is processed too.

As `todo-backend` doesn't have any submodules, we are not hitting this here so this is just thinking out loud about general approach.

I guess the latter is more convenient for development as you rebuild quickstart-parent so one less snapshot to deal with...

todo-backend/todo-backend.yaml
jbliznak: I understand, for testing purposes this was ok (and really one of the few ways how to actually test it), I just noted that because this should not be there in the time of merging

todo-backend/README.adoc
```diff
@@ -0,0 +1,304 @@
...
+The `todo-backend` quickstart demonstrates how to implement a backend that exposes a HTTP API with JAX-RS
+to manage a list of ToDo which are persisted in a database with JPA.
+
+This quickstart shows how to setup a local deployment of this backend as well as a deployment on OpenShift to connect
+to a PostgreSQL database also hosted on OpenShift.
+
+
+== What is it?
+
+The `todo-back` quickstart demonstrates how to implement a backend that exposes a HTTP API with `JAX-RS`
```
jbliznak, line 22: typo `todo-back`

README.adoc
```diff
@@ -504,6 +504,7 @@ NOTE: Some of these quickstarts use the H2 database included with WildFly. It is
 | link:tasks-rs/README{outfilesuffix}[tasks-rs]|JPA, JAX-RS | The `tasks-rs` quickstart demonstrates how to implement a JAX-RS service that uses JPA persistence. | Intermediate | _none_
 | link:temperature-converter/README{outfilesuffix}[temperature-converter]|CDI, JSF, SLSB EJB | The `temperature-converter` quickstart does temperature conversion using an EJB Stateless Session Bean (SLSB), CDI, and a JSF front-end client. | Beginner | _none_
 | link:thread-racing/README{outfilesuffix}[thread-racing]|Batch, CDI, EE Concurrency, JAX-RS, JMS, JPA, JSON, Web Sockets | A thread racing web application that demonstrates technologies introduced or updated in the latest Jakarta EE specification. | Beginner | _none_
+| link:todo-backend/README{outfilesuffix}[todo-backend]|JPA, JAX-RS, OpenShift, Galleon | The `todo-backend` quickstart demonstrates how to deploy a backend on OpenShift that exposes a HTTP API with JAX-RS and persists data in an external database with JPA | Intermediate | _none_
```
emmartins, line 507: no need for this, this table is automatically refreshed on build

pom.xml
emmartins, line 83: can you please move this to the quickstart pom.xml, as microprofile Quickstarts have

emmartins (changes requested):

pom.xml
jmesnil: Why must each quickstarts that uses bootable jar redefines this version?
How is it different from quickstarts sharing the same `version.microprofile.bom`?

todo-backend/todo-backend.yaml
jmesnil: you were right: let's have the PR being "clean" and do local changes for testing

todo-backend/todo-backend.yaml
jmesnil: during development, that is the only way to be able to build the image from a branch (as there are no released pushed to Maven repositories).
If I were using context-dir, I would not be able to verify that this Helm Chart works before the quickstarts have been released...

it's already been discussed and right now, it's not possible to build an Quickstart  image from a branch by pointing to the submodule directly.
To make it work, the quickstart MUST not have snapshots dependencies to WildFLy server and BOM.

jmesnil: it makes OpenShift development of the images more complex that it should really be...

jbliznak: yes, that's the drawback of using maven in general...

We are dealing with this by using maven repo proxy where we pre-upload all needed snapshot versions but I agree, it is not convenient

README.adoc
jmesnil: fixed, thanks

todo-backend/todo-backend.yaml
jmesnil: there is no drawback of using maven here, it's our design that causes issues:
* we rely on SNAPSHOTs of WildFly so we can not built the quickstart from the main branch
* we rely on a parent pom to get the quickstart dependencies so that it's not possible to `context-dir` a single quickstart and have it build in isolation

pom.xml
jmesnil: I'll revert this change but I don't understand the guiding principles here. When do we share things between quickstarts?

emmartins: At some point BOMs will move to each quickstart too, the future direction is that the parent is just an aggregator, and each QS is as much independent as possible. There are always issues due to the parent dependency, for instance wrt s2i testing and codeready workspaces.

jmesnil: ok, thanks for the explanation

emmartins (approved):

senatov (approved):

senatov (approved):

senatov (changes requested):
OK

senatov (approved):

todo-backend/README.adoc
```diff
@@ -0,0 +1,452 @@
...
+endif::[]
+
+=== Build and Deploy the Backend on OpenShift with {productName} S2I
+
+The backend will be built and deployed on OpenShift with a Helm Chart for {productName}.
+
+ifndef::ProductRelease[]
+[source,options="nowrap"]
+----
+$ helm install todo-backend -f todo-backend-s2i.yaml wildfly/wildfly
```
jbliznak, line 323: when user reads and tries the README from top down then this command will fail as the chart `todo-backend` is already installed via previous steps
```
Error: cannot re-use a name that is still in use
```

We can refer to the section `Remove the Backend` to clean it first

todo-backend/README.adoc
```diff
@@ -0,0 +1,452 @@
...
+----
+endif::[]
+
+The Helm Chart for this quickstart contains all the information to build an image from the source code using S2I:
+
+ifndef::ProductRelease[]
+[source,options="nowrap"]
+----
+build:
+  uri: https://github.com/wildfly/quickstart.git
```
jbliznak, line 347: @emmartins @jmesnil not sure if we want to have it hardcoded here, there is for sure a property for this which would make these  to `ifndef::ProductRelease[]` + `ifdef::ProductRelease[]` redundant

Note: it will still be hardcoded in the `todo-backend-s2i.yaml` though.

todo-backend/README.adoc
```diff
@@ -0,0 +1,452 @@
...
+  env:
+    - name: POSTGRESQL_PASSWORD
+      valueFrom:
+        secretKeyRef:
+          key: database-password
+          name: todos-db
+----
+
+When the application is deployed, the value for the `POSTGRESQL_PASSWORD` will be taken from the key `database-password`
+in the secrets `todos-db`.
```
jbliznak, line 382: probably a typo `secrets`

todo-backend/README.adoc
```diff
@@ -0,0 +1,452 @@
...
+
+Let's wait for the application to be built and deployed:
+
+[source,options="nowrap"]
+----
+$ oc get dc/todo-backend -w
```
jbliznak, line 388: This fails for me:
```
$ oc get dc/todo-backend -w
Error from server (NotFound): deploymentconfigs.apps.openshift.io "todo-backend" not found
```

There is no `todo-backend` DeploymentConfig created, I can only see `todo-backend` Deployment instead
```
$ oc get deployment/todo-backend -w
NAME           READY   UP-TO-DATE   AVAILABLE   AGE
todo-backend   0/3     3            0           31s
todo-backend   1/3     3            1           4m31s
todo-backend   2/3     3            2           4m31s
todo-backend   3/3     3            3           4m31s
```

I am not sure why the difference and how you managed to get the output listed, we are using Openshift 4.7 for testing

I can also see similar output for ReplicaSet, but here the name has generated suffix, no easy way to determine it so it is probably not useful:
```
$ oc get rs/todo-backend-5b649477b7 -w
NAME                      DESIRED   CURRENT   READY   AGE
todo-backend-5b649477b7   3         3         3       2m3s
```

todo-backend/README.adoc
```diff
@@ -0,0 +1,452 @@
...
+[source,options="nowrap"]
+----
+$ oc get route todo-backend -o jsonpath="{.spec.host}"
+todo-backend-jmesnil1-dev.apps.sandbox.x8i5.p1.openshiftapps.com
+----
+
+This value will be different for every installation of the backend.
+
+[WARNING]
+====
```
jbliznak, line 409: I would add here also note that the cluster we deployed on _must_ be publicly accessible (eg. we can't test it as our test clusters are limited to company network)

In such a case, we could redirect the user to test it the same way using `curl` we do in bootable deployment (but that section will not be there for product :-1: )

todo-backend/todo-backend-bootable-jar.yaml
```diff
@@ -0,0 +1,43 @@
+# This configuration file can be used to build and deploy the todo-backend
+# quickstart on OpenShift with the Helm Chart for WildFly.
+build:
+  uri: https://github.com/wildfly/quickstart.git
+  ref: todo-backend
```
jbliznak, line 5: this is a wrong branch

setting it right might be tricky though :( it appears we would need to change it for each release

todo-backend/todo-backend-s2i.yaml
```diff
@@ -0,0 +1,78 @@
+# This configuration file can be used to build and deploy the todo-backend
+# quickstart on OpenShift with the Helm Chart for WildFly.
+build:
+  uri: https://github.com/wildfly/quickstart.git
+  ref: todo-backend
```
jbliznak, line 5: this is a wrong branch, the same as for the previous yaml

todo-backend/README.adoc
```diff
@@ -0,0 +1,452 @@
...
+Once you have installed Helm, you need to add the repository that provides Helm Charts for {productName}:
+
+ifndef::ProductRelease[]
+[source,options="nowrap"]
+----
+$ helm repo add wildfly https://docs.wildfly.org/wildfly-charts/
+"wildfly" has been added to your repositories
+$ helm search repo wildfly
+NAME                    CHART VERSION   APP VERSION     DESCRIPTION
+wildfly/wildfly         1.1.0           22.0            Build and Deploy WildFly applications on OpenShift
```
jbliznak, line 264: I know this is only an example of the output but what about to parameterize the versions here?
Although I don't know how that works and will work for future WFs, there seems to be no branches on that repo

todo-backend/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+                            </feature-packs>
+                            <cloud/>
+                            <layers>
+                                <layer>cloud-server</layer>
+                                <layer>postgresql-datasource</layer>
+                                <layer>ejb</layer>
+                            </layers>
+                            <cli-sessions>
+                                <cli-session>
+                                    <!-- do not resolve expression as the CORS_ORIGIN en var -->
```
jbliznak, line 112: typo `en`

jbliznak (changes requested):

todo-backend/README.adoc
```diff
@@ -0,0 +1,452 @@
...
+* `POSTGRESQL_DATABASE` - the name of the database (that will be called `todos`)
+* `POSTGRESQL_SERVICE_HOST` - the host to connect to the database
+* `POSTGRESQL_SERVICE_PORT` - The port to connecto to the database
+* `POSTGRESQL_USER` & `POSTGRESQL_PASSWORD` - the credentials to connect to the database
+* `POSTGRESQL_DATASOURCE` - The name of the datasourcs (as mentioned above, it will be `ToDos`)
+
+=== Filters for Cross-Origin Resource Sharing (CORS)
+
+The Web frontend for this quickstart uses JavaScript calls to query the backend's HTTP API.
+We must enable Cross-Origin Resource Sharing (CORS) filters in the `undertow` subsystem of {productName}.
```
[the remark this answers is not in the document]
jmesnil: I reworded the sentence to make it clear that we need CORS to all HTTP request from Javascript to succeed.

CORS filters are not required when using curl.

todo-backend/README.adoc
jmesnil: I've added ellipsis for the command output (`...`), the version are not meaningful, it just shows how to install the chart repository.

todo-backend/README.adoc
jmesnil: that's fine. The user is not supposed to install both S2I and bootable jar examples at the same time.

jmesnil: and if if wants to install 2 applications, he just have to specify another name

todo-backend/README.adoc
jmesnil: the content of the YAML file will be different in EAP as it will require an additional env var to provision the datasource feature pack.

todo-backend/README.adoc
jmesnil: good catch, the helm chart was using deployment config when I wrote the initial readme but it is not using a `deployment`. I used your output to fix it.

todo-backend/README.adoc
```diff
@@ -0,0 +1,452 @@
...
+[source,options="nowrap"]
+----
+$ helm install todo-backend -f todo-backend-bootable-jar.yaml wildfly/wildfly
+NAME: todo-backend
+...
+STATUS: deployed
+REVISION: 1
+----
+
+The Helm Chart for this quickstart contains all the information to build an image from the source code using Bootable Jar:
```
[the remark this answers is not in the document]
jmesnil: the default branch will be used.

This looks like a maintenance nightmare as the quickstarts are never buildable from `master` or any other branch :(
We must specify the `ref` to correspond to the release tag every time.

jbliznak (approved):
Thanks for the updates

PR #486, opened 2021-03-29
[WFLY-14650] Warn message SRMSG18216: No `group.id` set during microp…

jstourac:
…rofile-reactive-messaging-kafka QS

https://issues.redhat.com/browse/WFLY-14650


emmartins:
@kabir please apply same change to 23.x, since that's the base of xp3

PR #487, opened 2021-03-29
[WFLY-11794] Fixing helloworld class file transformer quickstart example

ropalka:
https://issues.redhat.com/browse/WFLY-11794


ropalka:
@emmartins please review

emmartins:
@ropalka the client exec:java fails at my end with JDK12, I added a commit to fix the pom.xml dependencies and plugin config (plus some README enhancements) at #488

emmartins:
replaced by #488

PR #490, opened 2021-03-31
[WFLY-14651] Pre create topic to avoid LEADER_NOT_AVAILABLE

kabir:
https://issues.redhat.com/browse/WFLY-14651


kabir:
@emmartins I don't think the failure is related to my change

emmartins (approved):

PR #495, opened 2021-04-21
[WFLY-14702] Quickstart ejb-multi-server failing on client invocation…

jbaesner:
Issue: https://issues.redhat.com/browse/WFLY-14702

- fixed quickstart by adding the missing @SecurityDomain("other")
  annotation on the SLSBs
  This fixes the issue of the principal 'anonymous' seen in the output
- added the workaround to provide a default Elytron
  'authentication-context' for the transaction handling

### DIFF
--- a/ejb-multi-server/app-main/ejb/src/main/java/org/jboss/as/quickstarts/ejb/multi/server/app/MainAppBean.java
+++ b/ejb-multi-server/app-main/ejb/src/main/java/org/jboss/as/quickstarts/ejb/multi/server/app/MainAppBean.java
@@ -28,6 +28,8 @@ import javax.naming.Context;
 import javax.naming.InitialContext;
 import javax.naming.NamingException;
 
+import org.jboss.ejb3.annotation.SecurityDomain;
+
 import org.jboss.logging.Logger;
 
 /**
@@ -38,6 +40,7 @@ import org.jboss.logging.Logger;
  * @author <a href="mailto:wfink@redhat.com">Wolf-Dieter Fink</a>
  */
 @Stateless
+@SecurityDomain("other")
 public class MainAppBean implements MainApp {
     private static final Logger LOGGER = Logger.getLogger(MainAppBean.class);
     @Resource
--- a/ejb-multi-server/install-domain.cli
+++ b/ejb-multi-server/install-domain.cli
@@ -21,6 +21,7 @@ batch
 /host=master/core-service=management/security-realm=ApplicationRealm/authentication=local:remove
 
 /profile=default/subsystem=ejb3:write-attribute(name=default-missing-method-permissions-deny-access, value=false)
+/profile=full/subsystem=ejb3:write-attribute(name=default-missing-method-permissions-deny-access, value=false)
 /profile=ha/subsystem=ejb3:write-attribute(name=default-missing-method-permissions-deny-access, value=false)
 
 # Set the property for unique Tx node identifier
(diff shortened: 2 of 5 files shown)


emmartins:
@rachmatowicz can you please review?

rachmatowicz:
> @rachmatowicz can you please review?

Yes, will do it this afternoon.

ejb-multi-server/app-main/ejb/src/main/java/org/jboss/as/quickstarts/ejb/multi/server/app/MainAppBean.java
```diff
@@ -38,6 +40,7 @@
  * @author <a href="mailto:wfink@redhat.com">Wolf-Dieter Fink</a>
  */
 @Stateless
+@SecurityDomain("other")
```
rachmatowicz, line 43: This change obviously works and was present on the beans in the app-twoA/app-twoB servers. What is puzzling is that the ejb3 subsystem has an attribute called "default-security-domain" with value "other" and according to the definition of the attribute, it should set the security domain of beans without a security domain to "other" - which is effectively what this annotation is doing. I suspect that there is something fishy with the operation of this attribute.

ejb-multi-server/install-domain.cli
```diff
@@ -47,6 +48,11 @@ batch
 # can not commit a transaction after executing the SLSB as the bean (with context) is destroyed before !!
 /profile=full/subsystem=ejb3:write-attribute(name=default-slsb-instance-pool, value=slsb-strict-max-pool)
 
+# configure default transaction workaround
```
rachmatowicz, line 51: There is another workaround to this problem and that is getting the servers in domain mode to bind to "localhost" instead of 127.0.0.1. In that case, the client mappings entries will contain host=localhost and the problem will disappear. Might be worth considering.

emmartins:
@rachmatowicz what's the status of this, not clear to me, are you requesting changes to the PR or should this be merged?

rachmatowicz:
> @rachmatowicz what's the status of this, not clear to me, are you requesting changes to the PR or should this be merged?

@emmartins Requesting changes. Adding the system property to set jboss.bind.address and jboss.bind.address.management is a more direct fix to the underlying issue than to add a default security domain to the configuration.

rachmatowicz (changes requested):
I think this pull request looks good once the change to the bind addresses is made.

emmartins:
@jbaesner can you please address @rachmatowicz concerns?

ejb-multi-server/app-main/ejb/src/main/java/org/jboss/as/quickstarts/ejb/multi/server/app/MainAppBean.java
jbaesner: Ok, so I'll keep that introduced definition

ejb-multi-server/install-domain.cli
jbaesner: Instead of adding system properties, I'm going to update all interface definitions in host.xml and domain.xml using the install-domain.cli and replace the 127.0.0.1 values with localhost

jbaesner:
@emmartins, @rachmatowicz I pushed new changes. Now the Elytron default-authentication-context is being removed, but all interfaces are changed from default '127.0.0.1' to 'localhost'

chengfang (approved):

PR #500, opened 2021-06-07
[WFLY-14827] Add an explicit FacesConfiguration class

fischey:
This will fix https://issues.redhat.com/browse/WFLY-14827.

Currently it is necessary to either write an external Injection Point per JSF-specific injections points like @FacesContext or similiar.
Or you can use a class which is annotated with @FacesConfig which will give you the expected behaviour like in the docs.

In this PR a dedicated configuration class is used to activate JSF 2.3 features in JBoss EAP 7.3
Injection points like FacesContext and ExternalContext work out of the box in JBoss EAP 7.3 with this class.

No manual CDI Produces are necessary and JSF 2.3 works like described in the JSF docs.


emmartins:
@fischey thank you!
@fjuma can you please review?

fjuma (approved):
Thanks @fischey!

fischey:
Your welcome :-)

PR #505, opened 2021-06-29
[microprofile-config] Run Arquillian tests with Bootable Jar

jmesnil:
Switch to the arquillian bootable jar container to run the integration
tests. As the bootable jar contains the actual deployment, the tests
becomes a pure client tests (no @Deployment in the tests)

Tests can be run with `mvn clean verify -Pbootable-jar`

### DIFF
--- a/microprofile-config/pom.xml
+++ b/microprofile-config/pom.xml
@@ -22,6 +22,7 @@
     <maven.compiler.target>1.8</maven.compiler.target>
       <version.server.bootable-jar>24.0.0.Final</version.server.bootable-jar>
       <version.wildfly-jar.maven.plugin>5.0.0.Final</version.wildfly-jar.maven.plugin>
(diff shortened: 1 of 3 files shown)


jmesnil:
@xstefank  fiy, this is a draft PR to use the Bootable Jar Container to run the integration tests. 
IMHO, this is a better approach that testing against a remote WildFly server:

* it tests against the actual provisioned WildFly server in the bootable jar instead of a remote server that may/may not have the expected layers or features packs to run the quickstart
* the actual quickstart is deployed in the bootable jar: the tests becomes a pure test client that queries the actual endpoint from the application 

This PR modified the code to run the Arquillian tests with the Bootable Jar Container but I did not modify all the docs related to deploying and running the quickstarts.

xstefank:
@jmesnil great idea. I'll check it later and update other qs if this one is approved.

PR #521, opened 2021-09-16
[WFLY-15304] move quickstart's parent to its own repo

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-15304

This completes the move of the Quickstart's parent to its own repo, with its own release lifecycle.

For now I removed from the parent only the necessary to have its own release lifecycle, moving any dependency management to each quickstart pom.xml, in the future we may go further in such direction of having each quickstart as "complete" as possible, by keeping in the parent only the profiles and plugins related to the quickstart project (e.g. AsciiDoc build, wfly checkstyle, etc).

The original parent at the root of the quickstarts project now becomes just an aggregator, with only a single real child project, the "dist" one which builds the quickstarts zip release.

Related Parent PR: https://github.com/wildfly/wildfly-quickstart-parent/pull/1


emmartins:
@bstansberry @darranl any of you wanna have a look before I pull the trigger on this? :-)

emmartins:
Please note that the removal of maven properties on repositories specifications is due to maven properties resolving being done after resolving the parent. To keep such properties we would need to put the quickstarts parent in Maven Central, as we do with jboss-parent. 
ATM I don't think it worths the setup/delay of having quickstarts parent in Maven Central, I guess we can revisit that in the future if we ever have the need to override maven repos URLs through properties in quickstarts (yet there is always the SED alternative)

app-client/pom.xml
```diff
@@ -62,8 +60,7 @@
         <repository>
             <id>jboss-public-maven-repository</id>
             <name>JBoss Public Maven Repository</name>
-            <url>${maven.jboss.public.repository.url}</url>
-            <layout>default</layout>
+            <url>https://repository.jboss.org/nexus/content/groups/public/</url>
```
jmesnil, line 63: Can the `<repositories>` be moved to the parent too?

emmartins: @jmesnil you need the repos to at least retrieve the parent, which could be worked around by mirroring the parent to Maven Central. Moving there the repos also goes against the idea that each Quickstart should be as "complete" as possible, hiding in the parent only the stuff (profiles, plugins, etc) that are used for our project specifics (asciidoc, checkstyle, etc), since users would not use this parent.

emmartins: Please note that (when not using the parent) there are environments where we can't workaround and the user apps will need the repos, such as CodeReady Workspaces.

emmartins:
it seems the CI jobs are finally getting stable

darranl (approved):
Adding an approval, I think there has been some follow discussions raised but we can continue with those after Beta before Final.

jmesnil (approved):
🍾

emmartins:
thanks @darranl @jmesnil

PR #524, opened 2021-09-21
[WFLY-15323] add missing arquillian profiles

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-15323


emmartins:
@kabir please check if all good with mp reactive qs

PR #525, opened 2021-09-22
[WFLY-15089] Add startup health checks to microprofile-health quickstart

xstefank:
https://issues.redhat.com/browse/WFLY-15089

### DIFF
--- a/microprofile-health/README.adoc
+++ b/microprofile-health/README.adoc
@@ -23,8 +23,7 @@ or restarted.
 == Architecture
 
 In this quickstart, we have a simple REST application that exposes MicroProfile Health
-functionalities at the `/health/live` and `/health/ready` endpoints according to the
-specification.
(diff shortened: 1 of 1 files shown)


emmartins:
is this supposed to be just a README update?

xstefank:
@emmartins yes, I decided to not add just another (same) health check with just changed qualifier. We need to revisit this as @jmesnil was mentioning on one F2F. Too much of the same example with just a really small change (changed qualifier) IMO.

emmartins:
@xstefank so those new endpoints work without code changes?

xstefank:
@emmartins yes, liveness, readiness no changes. Only /health will aggregate everything including the new startup checks. But there are no needed adjustments in the quickstart for that. We test startup normally in test suite so I don't see much value in adding a code example here.

PR #527, opened 2021-10-04
[WFLY-15381] fix microprofile-faul-tolerance quickstart

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-15381

### DIFF
--- a/microprofile-fault-tolerance/pom.xml
+++ b/microprofile-fault-tolerance/pom.xml
@@ -149,8 +149,8 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>org.jboss.resteasy</groupId>
-            <artifactId>resteasy-jaxrs</artifactId>
+            <groupId>org.jboss.spec.javax.ws.rs</groupId>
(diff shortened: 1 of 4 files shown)


emmartins:
@jamezp wanna have a quick look? :-)

microprofile-fault-tolerance/pom.xml
```diff
@@ -150,7 +150,12 @@
         </dependency>
         <dependency>
             <groupId>org.jboss.resteasy</groupId>
-            <artifactId>resteasy-jaxrs</artifactId>
+            <artifactId>resteasy-core-spi</artifactId>
```
jamezp, line 153: It it's more critical to get the release out this isn't a huge deal, but this dependency can go away and in the `org.wildfly.quickstarts.microprofile.faulttolerance.Coffee.coffeeDetail()` the `@PathParam` import can change to `javax.ws.rs.PathParam` which is more appropriate IMO.

emmartins: done, and works, thanks

jamezp (approved):

PR #545, opened 2022-01-07
MicroProfile Reactive Messaging Fixes

spriadka:


spriadka:
@kabir Can you have a look at this?

kabir:
@spriadka as long as this just affects jkube (which I think it looks like) I think it is fine. Please confirm and I can merge it

spriadka:
@kabir Confirming it only affects jkube

PR #548, opened 2022-02-20
Helloworld jaxrs

ronsigal:
I've

  1 turned quickstart jaxrs-client into something simpler that parallels helloworld, and
  2. renamed it helloworld-jaxrs.

Note. jaxrs-client has two test classes: 1) ContactsRestClient, and 2) ContactsRestClientIT, where the latter uses Arquillian and the former doesn't. I'm not sure why they're both there (something to do with openshift?), but I've replicated the duplication in helloworld-jaxrs.


emmartins:
With the introduction of helloworld-rs I don't think we should still change this one to an helloworld too

emmartins:
PS: hope all is great with you @ronsigal  :)

PR #564, opened 2022-04-13
WildFly requires Java 11 to compile.

jamezp:


jamezp:
Actually this could be wrong. We may need to be testing against the 26.x branch really. What do you think @emmartins?

PR #574, opened 2022-06-22
[WFLY-16522] Download Kafka rather than docker-compose since docker-c…

kabir:
…ompose does not work with podman

https://issues.redhat.com/browse/WFLY-16522


zhantemirov (approved):
@emmartins please have a look at this, LGTM.

PR #578, opened 2022-08-11
[WFLY-16790] Jakarta EE 10 Migration (1st iteration)

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-16790


emmartins:
Please note that some of those quickstarts may have issues wrt Faces 4.0, which should be addressed as bugs by separated JIRAs, cause I lack the expertise for doing that now.

bstansberry (approved):

PR #589, opened 2022-10-25
[WFLY-17194] Verify and enable the the microprofile-metrics quickstart

jasondlee:
https://issues.redhat.com/browse/WFLY-17194

Enable microprofile-metrics module
Remove explicit protocol config so the correct default is used


bstansberry (approved):

bstansberry:
Thanks @jasondlee

PR #594, opened 2022-10-27
Update wildly-jar-maven-plugin to 8.0.1.Final

jmesnil:
for all quickstarts that uses Bootable Jar.


jmesnil:
Without that bump, the quickstarts that provides `bootable-jar` profiles fails with:

```
[ERROR] Failed to execute goal org.wildfly.plugins:wildfly-jar-maven-plugin:7.0.0.Final:package (default) on project microprofile-metrics: Provisioning failed:
Failed to parse /feature-pack.xml: ParseError at [row,col]:[3,143]
[ERROR] Message: Unexpected element '{urn:jboss:galleon:feature-pack:3.0}feature-pack'
[ERROR] -> [Help 1]
```

PR #625, opened 2022-12-21
correcting namespace faces-config.xml

kneringerjohann:
Namespace is incorrect on faces-config.xml


emmartins:
hi @kneringerjohann , sorry but this kind of fix/update is being worked out individually by the component leads related with each of the targeted quickstarts, so I am closing your PR. Thanks a lot tho.

PR #636, opened 2023-01-19
[WFLY-17399] Verify and re-enable ejb-txn-remote-call quickstart

gaol:
Issue: https://issues.redhat.com/browse/WFLY-17399

Verified on bare metal setup, I would like to enable the quickstart, and leave the verification on OpenShift as a new issue.

However, I did not find way yet to verify the steps running on Openshift platform, which has been migrated to use helm chart I believe.

Please reject it if the verification on OpenShift platform should be done before enabling it. thanks :)

### DIFF
--- a/ejb-txn-remote-call/server/extensions/postgresql-module.xml
+++ b/ejb-txn-remote-call/server/extensions/postgresql-module.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module name="org.postgresql.jdbc" xmlns="urn:jboss:module:1.8">
     <resources>
-        <artifact name="org.postgresql:postgresql:42.2.8"/>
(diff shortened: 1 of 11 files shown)


ejb-txn-remote-call/server/extensions/postgresql-module.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module name="org.postgresql.jdbc" xmlns="urn:jboss:module:1.8">
     <resources>
-        <artifact name="org.postgresql:postgresql:42.2.8"/>
+        <artifact name="org.postgresql:postgresql:42.2.27"/>
```
gaol, line 4: Updated the version because the old version will lead to Exception with Hibernate.

emmartins:
thanks @gaol , openshift stuff will be worked out on followup team task, no worries with that now

PR #660, opened 2023-04-19
[WFLY-17916] Disable MP Metrics and MP OpenTracing Quickstarts

jasondlee:
https://issues.redhat.com/browse/WFLY-17916

Remove the <module> entries for the two specs


bstansberry (approved):

emmartins:
@jasondlee we need to at least disable them on dist/src/main/assembly/assembly.xml 
@bstansberry shouldn't we also remove the related quickstart folders?

jasondlee:
> @jasondlee we need to at least disable them on dist/src/main/assembly/assembly.xml 
> @bstansberry shouldn't we also remove the related quickstart folders?

@emmartins, I thought I searched the project for references, but I guess I missed that one. I'll get it fixed. And, yes, I think we should remove those folder. I pinged on chat about that to sure about that first, but I'll remove them and update the PR

bstansberry (approved):

emmartins:
Merged, thanks.
PS: FYI there is no need to manually update the root README table of available quickstarts, such table is generated and updated automatically when doing a full project build.

PR #670, opened 2023-04-27
[WFLY-16273] add localization to kitchensink QS from the former kitch…

istudens:
…ensink-ml QS

Issue: https://issues.redhat.com/browse/WFLY-16273


emmartins:
sorry for the long delay, big thank you @istudens

PR #673, opened 2023-05-09
[WFLY-15167] added simple-webbapp-oidc quickstart

istudens:
Issue: https://issues.redhat.com/browse/WFLY-15167


pom.xml
```diff
@@ -345,6 +345,7 @@
                 <module>servlet-filterlistener</module>
                 <module>servlet-security</module>
                 <module>shopping-cart</module>
+                <module>simple-webapp-oidc</module>
```
rhusar, line 348: Not a fan of names like these for quickstarts; e.g. these are useless once sorted lexicographic. These should ideally follow naming such as `category-technology-scenario`, e.g. like `ha-singleton-deployment` or `microprofile-fault-tolerance`. So I would expect this to start with `oidc-...`.

rhusar (changes requested):

pom.xml
istudens: It's been renamed to oidc-webapp.
Thanks

PR #675, opened 2023-05-17
[WFLY-18027] moved package org.wildfly.quickstarts.todos to org.jboss…

istudens:
….as.quickstarts.todos

Issue: https://issues.redhat.com/browse/WFLY-18027


darranl:
Does this really need to change?  Across WildFly we already have many many packages already using org.wildfly

jamezp:
+1 I don't see a need for this at all. TBH if any of them are `org.jboss.*` they should be moved to `org.willdfly.*`. Not that we need to make that change, but that feels like the direction to go, not the opposite.

istudens:
This change is marked as critical for 8.0.0.Beta-CR1, see https://issues.redhat.com/browse/JBEAP-24264 .

darranl:
I think that is an issue that needs to be triaged further.  I am not sure it is valid but that needs to be discussed to come to a conclusion.

emmartins:
closing as agreed we wont do this change

PR #677, opened 2023-05-18
[WFLY-17949] improves CI jobs to cover provisioned-serv…

emmartins:
…er, bootable-jar and openshift profiles. The builds with deps are now also abstract to base branch, i.e. it checks out wildfly and boms branc with same name as the target branch of the PR.
Issue: https://issues.redhat.com/browse/WFLY-17949

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,8 +5,6 @@ name: WildFly Quickstarts CI
 on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-    branches:
-      - main
 jobs:
   Test-build-default-matrix:
     name: BUILD DEFAULT - JDK${{ matrix.jdk }} - ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -54,10 +67,12 @@ jobs:
     - uses: actions/checkout@v2
       with:
         repository: wildfly/wildfly
+        ref: ${{ github.base_ref }}
```
jamezp, line 70: One thing to note is this will require branches to exactly match. e.g. `main`, `28.x`. Each project would be required to have these branches.

emmartins:
@jfdenise I think this is now good to go
PS: the fails with deps are expected ( https://issues.redhat.com/browse/WFLY-18021 ), and actually a good indicator this is a nice improvement

.github/workflows/ci.yml
emmartins: yes, that's intended, BOMs and QS always follow same branch names as WildFly

jfdenise (approved):

PR #686, opened 2023-06-14
[WFLY-18127] remove opentelemetry layer exclusion workaround

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-18127


emmartins:
@ehsavoie please check if this is all good

PR #699, opened 2023-09-07
Update the CI runner

jamezp:
Also add dependabot for upgrading GitHub Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,11 +24,12 @@ jobs:
     - uses: actions/checkout@v4
       with:
         path: quickstarts
-    - name: Set up JDK ${{ matrix.java }}
-      uses: AdoptOpenJDK/install-jdk@v1
+    - name: Set up JDK ${{ matrix.jdk }}
+      uses: actions/setup-java@v3
       with:
-        version: ${{ matrix.jdk }}
-        impl: hotspot
+        java-version: ${{ matrix.jdk }}
+        distribution: 'temurin'
+        cache: 'maven'
     - name: Build Quickstarts provisioned-server profile
       run: |
         cd quickstarts
@@ -77,11 +78,12 @@ jobs:
     - uses: actions/checkout@v4
       with:
         path: quickstarts
-    - name: Set up JDK ${{ matrix.java }}
-      uses: AdoptOpenJDK/install-jdk@v1
+    - name: Set up JDK ${{ matrix.jdk }}
+      uses: actions/setup-java@v3
       with:
-        version: ${{ matrix.jdk }}
-        impl: hotspot
+        java-version: ${{ matrix.jdk }}
+        distribution: 'temurin'
+        cache: 'maven'
     - name: Build Server
       run: |
         cd wildfly
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         jdk: [11, 17]
         os: [ubuntu-20.04, windows-latest]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
       with:
         path: quickstarts
     - name: Set up JDK ${{ matrix.java }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,9 +3,24 @@
 name: WildFly Quickstarts CI
 
 on:
+  push:
+    branches:
+      - 'main'
```
emmartins, line 8: The workflow should be "branch abstract", to work on the NN.x branches without any modifications. This is why all dependency checkouts use ref: ${{ github.base_ref }}

.github/workflows/ci.yml
```diff
@@ -3,9 +3,24 @@
...
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+
+# Only run the latest job
+concurrency:
+  group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
+  cancel-in-progress: true
+
 jobs:
+  wildfly-build:
```
emmartins, line 18: The workflow should be "branch abstract", to work on the NN.x branches without any modifications. This is why all dependency checkouts use ref: ${{ github.base_ref }}

.github/workflows/ci.yml
```diff
@@ -55,47 +68,43 @@ jobs:
...
         repository: wildfly/boms
         ref: ${{ github.base_ref }}
         path: boms
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
       with:
         path: quickstarts
-    - name: Set up JDK ${{ matrix.java }}
-      uses: AdoptOpenJDK/install-jdk@v1
+    - uses: actions/download-artifact@v3
```
emmartins, line 91: What's the purpose of this artifact download?

.github/workflows/ci.yml
emmartins: To be clear, the "with deps" job of the 29.x branch of QS should build with WFLY 29.x and BOMs 29.x built with such WFLY.

.github/workflows/ci.yml
emmartins: Ok nevermind I saw in the WFLY CI workflow sources this is related with the reusable workflow. I should note that this makes the job look more complex / less readable than the simple checkout and mvn build tho ;-)

emmartins: By the way, is it possible to start the WFLY built and downloaded this way?

jamezp: There is no WildFly distribution built here, it's the maven repository which contains the dependencies required for provisioning a SNAPSHOT of WildFly for testing.

.github/workflows/ci.yml
jamezp: That only kind of works. What I mean by that, is that each branch will have it's own CI. Once you branch, e.g. 29.x, then that branch runs off it's own workflow, not the one from main.

That said, this change is not needed at this point really. It's really to get around dependabot jobs from running two different CI runs. That said, there are other ways to handle that.

.github/workflows/ci.yml
jamezp: Yes, we could change the `${{ github_base_ref}}` here for sure. I'll fix that.

jamezp:
I've pushed changes to not use `main` for any branch names.

jamezp:
There is not a dynamic way to use the branch for the reusing the wildfly-build. I've hard-coded it to main for now, but it would need a change once we branch. However, I do feel personally that is okay. If it's not something you want to do though then we can just ditch this PR.

IMO it's good to share builds in case things change there is only one place they need to change.

emmartins:
https://github.com/wildfly/quickstart/pull/696 was merged now and ci.yml is now project_ci.yml, and most of the enhancements here are there already. What's missing is the usage of WFLY workflow to build it, which I would like to further discuss it with you. Once you get back from PTO please ping me wrt that.

emmartins:
Closing since those enhancements are now all being used by the new CI workflows

PR #710, opened 2023-09-28
Fix for WFLY-18567, Upgrade cloud and datasources Galleon feature-packs in quickstarts

jfdenise:


emmartins:
thank you @jfdenise

PR #713, opened 2023-09-29
Kubernetes testing

jmesnil:
* Set up a Kubernetes cluster (using Kind) with an image registry and an ingress controller
* Build container images using wildfly:image goal and push them to the cluster image registry
* Add an ingress to access the application (until https://github.com/wildfly/wildfly-charts/issues/52 is resolved)

* Add tests for helloworld & todo-backend

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,142 @@
+name: Kubernetes Testing
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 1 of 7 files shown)


jmesnil:
@kabir @emmartins This PR is not meant to be merged but this shows how we could use a local Kubernetes cluster to test the QuickStarts. I did if for the simplest one (`helloworld`) and the most complex one (`todo-backend`).

It does not rely on S2I to build the application image and use the `wildfly:image` goal instead.

jmesnil:
You can see executions of the action at https://github.com/jmesnil/quickstart/actions/workflows/testing.yml

PR #714, opened 2023-09-29
Fix OpenShift deployment of todo-backend

jmesnil:
Its openshift profile must provision the Cloud feature pack to properly deploy on OpenShift.

Remove the CORS env vars from the chart as they are already set at provisioning time with the cors_filter.cli script


jmesnil:
@ehsavoie fiy

PR #718, opened 2023-10-04
WFLY-18474: helloworld-mdb Quickstart Common Enhancements CY2023Q3

ehsavoie:
* Updating the docs
 * Adding an integration test
 * Updating the pom

### DIFF
--- a/helloworld-mdb/pom.xml
+++ b/helloworld-mdb/pom.xml
@@ -44,11 +44,13 @@
     </licenses>
 
     <properties>
-        <!-- The versions for BOMs, Dependencies and Plugins -->
-        <version.server.bom>30.0.0.Beta1</version.server.bom>
+        <!-- the version for the Server -->
         <version.server>30.0.0.Beta1</version.server>
-        <version.wildfly.maven.plugin>4.2.0.Final</version.wildfly.maven.plugin>
-        <version.cloud.fp>5.0.0.Beta1</version.cloud.fp>
+        <!-- The versions for BOMs, Packs and Plugins -->
+        <version.bom.ee>${version.server}</version.bom.ee>      
+        <version.pack.cloud>5.0.0.Beta1</version.pack.cloud>
+        <version.plugin.wildfly>4.2.0.Final</version.plugin.wildfly>
+        <version.junit-jupiter-engine>5.10.0</version.junit-jupiter-engine>
     </properties>
 
     <repositories>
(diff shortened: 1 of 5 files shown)


helloworld-mdb/pom.xml
```diff
@@ -147,6 +148,12 @@
             <artifactId>jakarta.annotation-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <version>5.10.0</version>
```
emmartins, line 154: IMHO There is no value in having to manage Junit Jupiter version on each QS *now*, I think it would be preferable to now use standard junit like specified by the doc, which version comes managed from the EE with Tools BOM. I think it's better to re-evaluate Junit5 later as I told you, and if we go for such change we then add it's version management in the EE With Tools BOM. Still, if you want to proceed with Junit5 now you should define it's version in the properties like this:


<version.junit-jupiter-engine>5.10.0</version.junit-jupiter-engine>

... and then refer such property in dependencyManagement or directly in the dependency.

helloworld-mdb/src/test/java/org/wildfly/quickstarts/mdb/BasicRuntimeIT.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * The very basic runtime integration testing.
+ *
+ * @author Emmanuel Hugonnet
+ * @author emartins
+ */
+public class BasicRuntimeIT {
+
+    private static final String DEFAULT_SERVER_HOST = "http://localhost:8080";
```
emmartins, line 39: private static final String DEFAULT_SERVER_HOST = "http://localhost:8080/helloworld-mdb";

emmartins (changes requested):
Besides the changes required (BasicRuntimeIT) and suggested (drop to Junit) there is something missing, which is the .github/workflows/quickstart_helloworld-mdb_ci.yml file providing the quickstart's CI, which should be similar to https://github.com/wildfly/quickstart/blob/main/.github/workflows/quickstart_helloworld_ci.yml (just add -mdb to all helloworld ocurrences)

Other than that I believe at some point we will simply move the clustering content to the new app development guide in the works, but for now let's keep it as it is (no integration testing needed)

helloworld-mdb/pom.xml
ehsavoie: I agree with @jamezp we should move to JUnit 5

emmartins (approved):

PR #744, opened 2023-10-14
[WFLY-18502][WFLY-18498] adds MATRIX_OS allowing customization of tes…

emmartins:
…t matrix OSes; adds support for optional quickstart-specific sh scripts to run before and after the tests"

### DIFF
--- a/.github/workflows/quickstart_ci.yml
+++ b/.github/workflows/quickstart_ci.yml
@@ -22,6 +22,11 @@ on:
         required: false
         default: true
         type: boolean
+      MATRIX_OS:
+        description: 'the OSes to be used on the test matrix, i.e. matrix.os'
+        required: false
+        default: '"ubuntu-latest", "windows-latest"'
+        type: string
       SERVER_PROVISIONING_SERVER_HOST:
         description: 'the server.host arg to be used when testing server provisioning with the mvn verify command'
         required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/quickstart_ci.yml
```diff
@@ -52,6 +66,16 @@ jobs:
           java-version: ${{ matrix.jdk }}
           distribution: 'temurin'
           cache: 'maven'
+      - name: Run before script
+        env:
+          FILE: "./quickstarts/.github/workflows/quickstart_${{ inputs.QUICKSTART_PATH }}_ci_before.sh"
```
kabir, line 71: Maybe but this one and the other file as the env for the whole job to reduce duplication? IMO having it at the top level also makes it more 'documented'

.github/workflows/quickstart_ci.yml
```diff
@@ -93,11 +117,21 @@ jobs:
...
+              chmod +x $FILE 
+              bash $FILE
+          fi
+        shell: bash
       - uses: actions/upload-artifact@v3
         if: failure()
         with:
           name: surefire-reports-JDK${{ matrix.jdk }}-${{ matrix.os }}
-          path: 'quickstarts/$QUICKSTART_PATH/**/surefire-reports/*.txt'
+          path: 'quickstarts/${{ env.QUICKSTART_PATH }}/**/surefire-reports/*.txt'
```
kabir, line 134: I **_think_** this should be `${{ inputs.QUICKSTART_PATH }}` but am not 100% sure!

ehsavoie: +1

.github/workflows/quickstart_ci.yml
emmartins: My idea was to have enforce each QS stays independent .yml and .sh wise. Duplication should be rare, if even will exist, and I will of course document this is at the google doc.

PR #765, opened 2023-10-24
WFLY-18473: helloworld-jms Quickstart Common Enhancements CY2023Q3.

ehsavoie:
* Using integration-testing as a way to execute the client
* provisionned server is ready to handle the client

Issue: https://issues.redhat.com/browse/WFLY-18473


helloworld-jms/README.adoc
```diff
@@ -117,9 +102,9 @@ timestamp org.jboss.as.quickstarts.jms.HelloWorldJMSClient main
 INFO: Received message with content Hello, World!
 ----
 
-== Optional Properties
+=== Optional Properties
```
emmartins, line 105: Personally I would get ridden of those (here and in the pom.xml) and hardcode them in the test class, like other QS are doing. The only non constant would be the java.naming.provider.url but you could derivate the value for that one from the server.host param used in tests

helloworld-jms/src/test/resources/helloworld-jms-logging.properties
```diff
@@ -0,0 +1,39 @@
+#
```
emmartins, line 1: do we really need logging properties on a helloworld? This quickstart should be as simple as possible, focusing on JMS only)

helloworld-jms/src/main/resources/helloworld-jms-logging.properties
```diff
@@ -0,0 +1,39 @@
+#
```
emmartins, line 1: do we really need logging properties on a helloworld? This quickstart should be as simple as possible, focusing on JMS only)

helloworld-jms/README.adoc
```diff
@@ -167,32 +152,4 @@ This script removes the `test` queue from the `messaging` subsystem in the serve
...
-// Product Release content only
-//*************************************************
-ifdef::ProductRelease[]
-
-// Quickstart not compatible with OpenShift
-include::../shared-doc/openshift-incompatibility.adoc[leveloffset=+1]
-
-endif::[]
\ No newline at end of file
+include::../shared-doc/openshift-incompatibility.adoc[leveloffset=+1]
```
emmartins, line 155: Before this please include the provisioning instructions:

```
// Build and run sections for other environments/builds
ifndef::ProductRelease,EAPXPRelease[]
include::../shared-doc/build-and-run-the-quickstart-with-provisioned-server.adoc[leveloffset=+1]
endif::[]
```

emmartins:
In generally I think this quickstart could be much simplified, by removing the optional properties (which are not used anywhere in the user flow), and also the logging properties. This would mean everything non standard on it would focus on JMS only. Those changes are optional of course, just suggestions, the only mandatory change needed is to include the provisioning instructions.

emmartins (changes requested):
In generally I think this quickstart could be much simplified, by removing the optional properties (which are not used anywhere in the user flow), and also the logging properties. This would mean everything non standard on it would focus on JMS only. Those changes are optional of course, just suggestions, the only *mandatory* change needed is to include the provisioning instructions.

helloworld-jms/README.adoc
ehsavoie: I put them here for retro compatibility

PR #772, opened 2023-11-06
[WFLY-18465] ejb-remote Quickstart Common Enhancements CY2023Q3

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-18465

Besides the changes specified by CY2023Q3 enhancements, this also includes the following major changes:
* Changes maven project to single module war project, moving server side component sources to /src
* Updates README considering the maven project changes
* Adds root index.html, which may be used to check successful deployment
* Moves client code to EJBRemoteIT.java integration testing


tadamski (approved):
Double checking with the document it seems that the required changed are present, and from EJB perspective iiuc there are no major changes, so for me this should be ready to go...

PR #773, opened 2023-11-10
WFLY-18488 jts Quickstart Common Enhancements CY2023Q3

mmusgrov:
https://issues.redhat.com/browse/WFLY-18488

The changes include clarifying the test setup and disabling testing on Openshift since setting up and coordinating multiple servers is not technically feasible without incurring a lot of effort.

### DIFF
--- a/.github/workflows/quickstart_jts_ci.yml
+++ b/.github/workflows/quickstart_jts_ci.yml
@@ -1,0 +1,16 @@
+name: WildFly jts Quickstart CI
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - jts/**'
+      - .github/workflows/quickstart_ci.yml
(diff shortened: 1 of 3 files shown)


emmartins (changes requested):
Hi @mmusgrov , considering the QS complexity I agree there is no point (at least now) in targeting OpenShift compatibility, and I believe the same may be said about server provisioning. Still there are are few things missing:
* the CI yaml for this quickstart, which should specify the optional params TEST_OPENSHIFT and TEST_PROVISIONED_SERVER as false
* wildfly plugin management in the pom.xml

mmusgrov:
Ah, I had those but forgot to git add the new file, sorry about that, I will update the PR.

emmartins (approved):

emmartins:
thank you @mmusgrov

PR #775, opened 2023-11-15
[WFLY-18483] Update the jaxrs-jwt for the Common Enhancements. Make t…

jamezp:
…his quickstart a single module and use the test for the client side example.

https://issues.redhat.com/browse/WFLY-18483


jamezp:
It looks like there is a race in the reload, which makes sense. I've filed https://issues.redhat.com/browse/WFMP-224 to address some reload cases in the plugin that should help. Once the rest of the tests run and I have a chance to test with OpenShift, I'll fix that reload.

emmartins:
FYI other QS did remove the reload (it doesn't make much sense for provisioning anyway), and I reported this previously at https://issues.redhat.com/browse/WFLY-18698

jamezp:
> FYI other QS did remove the reload (it doesn't make much sense for provisioning anyway), and I reported this previously at https://issues.redhat.com/browse/WFLY-18698

It makes sense for other goals like `start`, `dev` and `run` though. I agree it definitely doesn't make sense for the packaging and provisioning goals though. I really think this needs to be done in the `wildfly-maven-plugin` which I'll likely do today.

jaxrs-jwt/pom.xml
```diff
@@ -30,29 +30,243 @@
...
+    </dependencies>
+
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.wildfly.plugins</groupId>
+                    <artifactId>wildfly-maven-plugin</artifactId>
+                    <version>${version.plugin.wildfly}</version>
+                    <configuration>
```
emmartins, line 143: wild guess but probably this that is failing server provisioning testing, perhaps revert to just configure versions like others?

jamezp: It's a bug in the plugin. There is a simple workaround for it, but I'll file an issue too.

jamezp: For reference https://issues.redhat.com/browse/WFMP-227.

emmartins:
thank you @jamezp

PR #777, opened 2023-11-17
remove unnecessary resteasy spring feature pack configuration

liweinan:
The `feature pack` configuration is unnecessary because it's already included in WildFly.


liweinan:
I have verified the change locally:

```bash
➤ mvn wildfly:run -Pprovisioned_server
```

```bash
➤ curl 'http://localhost:8080/spring-resteasy/hello?name=yourname'
Welcome to RESTEasy + Spring, yourname.⏎
```

PR #783, opened 2023-11-27
[WFLY-18490] logging README minor fixes

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-18490

A follow up of #779:
- adds attribute that identifies the qs as openshift compatible (in the root README table of available QS) 
- fixes the order of sections, moving provisioning and openshift to bottom
- removes deprecated sections (codeready, debug)
- fixes include of standard dist testing adoc, adding a line break before it, and correcting the offset


jamezp (approved):
Sorry about missing the `:openshift: true` part. Thanks for moving the other stuff to the correct place :)

PR #780, opened 2023-11-22
[WFLY-18499] Update the microprofile-rest-client for the Common Enhan…

jamezp:
…cements. Refactor the quickstart to a single module with a test for the MicroProfile REST Client.

https://issues.redhat.com/browse/WFLY-18499

PR #789, opened 2023-11-30
[WFLY-18499] mp rest-client README rework

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-18499
Followup of #780, reworks the README:
- Removes from scratch content that was outdated
- Adds 2 sections wrt accessing the app, and wrt the rest client

### DIFF
--- a/microprofile-rest-client/README.adoc
+++ b/microprofile-rest-client/README.adoc
@@ -33,186 +33,23 @@ include::../shared-doc/use-of-jboss-home-name.adoc[leveloffset=+1]
 // Start the {productName} Standalone Server
 include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
 
-== Solution
-
-We recommend that you follow the instructions that
-<<creating-new-project, create the application step by step>>. However, you can
-also go right to the completed example which is available in this directory.
-
 // Build and Deploy the Quickstart
 include::../shared-doc/build-and-deploy-the-quickstart.adoc[leveloffset=+1]
-// Server Distribution Testing
-include::../shared-doc/run-integration-tests-with-server-distribution.adoc[leveloffset=+1]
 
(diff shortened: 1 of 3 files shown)


jamezp:
@emmartins Do we want to remove the scratch content? I thought it was pretty good and when I walked through it it worked for me locally?

The other changes look fine.

emmartins:
It didn’t work, and I believe this is what Yeray was reporting in the JIRA
too, cause it was missing some needed parts, but that actually wasn’t the
main reason for the rework, it was still targeting the old structure of
country server be country client, not matching your rework. Honestly
keeping up to date “from scratch” instructions have been a challenge to
maintain, and not just for this particular quickstart, but if you prefer to
keep it that way and fix the old content that’s fine. Personally I think
those are fine once, then become redundant, thus the “move” to the new app
dev doc.

jamezp:
My only argument for keeping it would be this is supposed to be a client example, and it's kind of not a client example. The test is, but I'm not sure how relevant that is. The walk through just shows how to create the actual client which is kind of the point of the quickstart.

emmartins:
@jamezp this was just my proposal for a quick fix, it would be good to have a fix for WFLY 31 beta, pls let me know if you want to propose something else

emmartins:
sorry, wrong button, unintentional close :-)

jamezp:
@emmartins It's up to you. I think they're good instructions, but if others tested and they didn't work I guess delete them. I'm not sure when I'd have time to update them TBH. I've got a lot on my plate right now.

emmartins:
@jamezp ok I will merge this now, since it makes it read for xp5 qe, and whenever you have a chance and will feel free to rework such doc.

PS: we should end up moving all those "from scratch" content and integrate in the new guide anyway

PR #791, opened 2023-12-01
[WFLY-18745] introduced new single-step profile for integration tests…

istudens:
… with provisioned server

Issue: https://issues.redhat.com/browse/WFLY-18781

### DIFF
--- a/helloworld/pom.xml
+++ b/helloworld/pom.xml
@@ -235,6 +235,50 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>integration-testing-with-provisioned-server</id>
+            <build>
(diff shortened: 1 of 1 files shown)


helloworld/pom.xml
```diff
@@ -235,6 +235,50 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>integration-testing-with-provisioned-server</id>
```
jamezp, line 239: We also need the `maven-failsafe-plugin` defined and a way to define the `server.host` by default IMO. This won't work unless you explicitly set that property if used with the `provisioned-server` profile.

PR #792, opened 2023-12-01
[WFLY-18814] fixed typo - server.host instead of jbossHome

istudens:
Issue: https://issues.redhat.com/browse/WFLY-18814


emmartins:
not a bug, please see JIRA

PR #794, opened 2023-12-04
[WFLY-18769] Infrastructure for publishing quickstart html content on…

bstansberry:
… GitHub pages

PART OF https://issues.redhat.com/browse/WFLY-18769

@emmartins This is the second from last commit from my https://github.com/bstansberry/quickstart/commits/WFLY-18769 branch; i.e. the one that doesn't touch any of the poms or README.adoc files in the actual QS modules.

### DIFF
--- a/.github/workflows/publish-pages.yml
+++ b/.github/workflows/publish-pages.yml
@@ -1,0 +1,67 @@
+name: Deploy Quickstart html to Pages
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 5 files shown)


quickstarts.zip
emmartins, line 1: unintentional I think

bstansberry: Yikes!

bstansberry: Fixed. Thanks @emmartins

PR #799, opened 2023-12-11
[WFLY-18468] ejb-throws-exception Quickstart Common Enhancements CY20…

tadamski:
…23Q3

https://issues.redhat.com/browse/WFLY-18468

This quickstart is OpenShift incompatible


emmartins (changes requested):
* quickstart_ejb-throws-exception_ci.yml missing at /.github/workflows (please note that since this is not openshift compatible it should define TEST_OPENSHIFT: false
* all README.adoc changes missing

tadamski:
fixed

PR #800, opened 2023-12-11
[WFLY-18769] move publish pages GH Workflow to JDK 17, needed by spri…

emmartins:
…ng-resteasy quickstart
Issue: https://issues.redhat.com/browse/WFLY-18769


emmartins:
@bstansberry I guess there is no issue changing this? The spring-resteasy QS is only built with JDK17+

bstansberry (approved):

PR #801, opened 2023-12-11
fix the variable names in the `batch-processing` example.

liweinan:
There are some typos in the example code.


liweinan:
The example is verified to be run locally without problem:

image 3

emmartins:
thanks

PR #819, opened 2023-12-15
[WFLY-18853] added pluginManagement section for wildfly-maven-plugin

istudens:
Issue: https://issues.redhat.com/browse/WFLY-18853


istudens:
@emmartins can you review this please? thanks

emmartins:
good catch, thanks

PR #844, opened 2024-01-05
 [WFLY-18887]: todo-backend quickstart CI test fails.

ehsavoie:
Allowing to pass extra argumnets when starting the server using the 'EXTRA_RUN_ARGS' parameter.

Issue: https://issues.redhat.com/browse/WFLY-18887


emmartins (changes requested):
all looks good but the typo should be fixed, and please rename the commit msg too

emmartins (approved):

PR #852, opened 2024-01-17
Add Glow plugin examples

kabir:


shared-doc/build-and-run-the-quickstart-with-bootable-jar.adoc
```diff
@@ -34,6 +35,51 @@ The quickstart `pom.xml` file contains a Maven profile named *bootable-jar* whic
...
+ifdef::portedToGlow[]
+[source,xml,subs="attributes+"]
+----
+      <profile>
+          <id>bootable-jar</id>
+          <build>
+              <plugins>
+                  <plugin>
+                      <groupId>org.wildfly.plugins</groupId>
+                      <artifactId>wildfly-jar-maven-plugin</artifactId>
```
jfdenise, line 48: should be wildfly-maven-plugin

jfdenise: Perhaps we need the version too 5.0.0.Beta2, the latest one is 4.2.x that doesn't support glow.

PR #854, opened 2024-01-22
 WFLY-18925: Flattened Quickstart READMEs

theashiot:
Issue: https://issues.redhat.com/browse/WFLY-18925


.github/workflows/reduce-readme.yml
```diff
@@ -0,0 +1,34 @@
+name: Reduce README
+on:
+  push:
+    paths:
```
emmartins, line 4: this and branch is probably not needed as currently configured, it's default value

.github/workflows/reduce-readme.yml
```diff
@@ -0,0 +1,34 @@
...
+    - '**'
+    branches: ['**']
+jobs:
```
emmartins, line 7: We also need to be able to cancel new runs of same job so please add:
```
# Only run the latest job
concurrency:
  group: '${{ github.workflow }} @ ${{ github.ref || github.run_id }}'
  cancel-in-progress: true
```

.github/workflows/reduce-readme.yml
```diff
@@ -0,0 +1,34 @@
+name: Reduce README
```
emmartins, line 1: Rename to "WildFly Quickstarts Reduce READMEs"

.github/workflows/reduce-readme.yml
```diff
@@ -0,0 +1,34 @@
...
+      uses: actions/checkout@v3
+    - name: Install Asciidoctor Reducer
+      run: sudo gem install asciidoctor-reducer
+    - name: Reduce README
+      run: |
+        CURRENT_DIR="$( pwd -P)"
+        #Get a list of subdirectories that don't start with a .
+        subdirs=`find . -maxdepth 1 -type d ! -iname ".*"`
+        #Loop over the subdirectories
+        for subdir in $subdirs
```
emmartins, line 21: The redution examples didn't resolve all attributes, for instance {artifactId} (which could be simply the subdir value) coming from qs parent, or others defined locally in quickstart pom.xmls. I believe this is an acceptable limitation, but if you want to look for ways to solve this may be handy, it shows how to obtain property values from pom.xml and then inject it on another mvn command:

```
        echo "VERSION_BOMS=$(mvn -N org.apache.maven.plugins:maven-help-plugin:3.2.0:evaluate -Dexpression=project.version -q -DforceStdout)" >> $GITHUB_ENV
        mvn -U -B -fae clean package -Pprovisioned-server -Dversion.server.bom=${{ env.VERSION_BOMS }}

```

emmartins (changes requested):
Please also remove the README.adoc

.github/workflows/reduce-readme.yml
theashiot: fixed

.github/workflows/reduce-readme.yml
theashiot: added

.github/workflows/reduce-readme.yml
theashiot: done

emmartins (approved):

PR #870, opened 2024-01-29
[WFLY-18943] Fix automated testing of OpenTelemetry QS under OpenShift

jasondlee:
https://issues.redhat.com/browse/WFLY-18943

Build with openshift profile
Clean up YAML
Modify how WF is configured under OS build
Fix application deployment in POM file


emmartins (commented):
Some changes are not needed, you may find the proposed changes at https://github.com/wildfly/quickstart/pull/872

PR #871, opened 2024-01-30
[WFLY-18974] - Re-enable OS testing for micrometer and otel QSs

jasondlee:
https://issues.redhat.com/browse/WFLY-18974

Add runPostHelmInstallCommands to configure otel collector
Remove exclude entries for these two QSs

(Includes changes from WFLY-18942 and WFLY-18943, but those will be resolved once those branches are merged)


jasondlee:
These changes have been incorporated into #872 and #873.

PR #885, opened 2024-03-25
[WFLY-18476] Additional fixes/enhancements

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-18476
Changes:
* Moves README.adoc to README-source.adoc
* Moves section set_up_client_keystore_using_java_keytool to after the server is started
* Adds extraStandardDistTestParams and extraProvisioningTestParams extension points to add extra testing params
* Removes include for outdated sections wrt IDEs and debug
* Syncs pom version properties
* Moves http-client dependency to manged version, and test scope
* Simplifies basic test usage of a server.dir property

### DIFF
--- a/helloworld-mutual-ssl/README-source.adoc
+++ b/helloworld-mutual-ssl/README-source.adoc
@@ -27,84 +27,22 @@ include::../shared-doc/system-requirements.adoc[leveloffset=+1]
 // Use of {jbossHomeName}
 include::../shared-doc/use-of-jboss-home-name.adoc[leveloffset=+1]
 
-[[set_up_client_keystore_using_java_keytool]]
-== Set Up the Client Keystore Using Java Keytool
-
-. Open a terminal and navigate to the {productName} server `configuration` directory:
-+
-[source,subs="+quotes,attributes+",options="nowrap"]
-----
-$ cd __{jbossHomeName}__/standalone/configuration/
-----
-
-. Create the client certificate, which is used to authenticate against the server when accessing a resource through TLS.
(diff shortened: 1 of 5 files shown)


PrarthonaPaul:
Hi @emmartins 
Thanks for adding the changes. 
I think they look good. One note: for the tests, we need to specify the server home using `-Djboss.server.config.dir=PATH/TO/WIDFLY/HOME/standalone/configuration` so the certificates can be found by the test. This is also the case when running the tests using a provisioned server.

emmartins:
Thanks @PrarthonaPaul , please note that I did change that a bit too, the property was renamed to "server.dir", and it should point only to the server base dir, the test than joins the "standalone/configuration" path to it. Also, it defaults to target/server, so no param is needed when testing server provisioning. 

The doc rebuilt is now complete and you now have this server.dir param in the standard server dist testing command (see https://github.com/wildfly/quickstart/tree/main/helloworld-mutual-ssl)

PrarthonaPaul:
> Thanks @PrarthonaPaul , please note that I did change that a bit too, the property was renamed to "server.dir", and it should point only to the server base dir, the test than joins the "standalone/configuration" path to it. Also, it defaults to target/server, so no param is needed when testing server provisioning.
> 
> The doc rebuilt is now complete and you now have this server.dir param in the standard server dist testing command (see https://github.com/wildfly/quickstart/tree/main/helloworld-mutual-ssl)

That sounds good! 
I was thinking the same too while adding the comment. Thank you for making the edits!

PR #891, opened 2024-04-08
Upgrade to WildFly 32 beta1 and WFMP 5.0.0.Beta5

jfdenise:


emmartins:
hi @jfdenise, failed to notice this one, but think it's no longer needed? We are in the process of merging  glow to main today...

PR #916, opened 2024-05-21
Remove the need to use a different feature-packs URL for WildFly SNAPSHOT

jfdenise:
Recent refactoring of https://github.com/wildfly/wildfly-galleon-feature-packs removed this requirement.
@kabir FYI.

### DIFF
--- a/.github/workflows/quickstart_ci.yml
+++ b/.github/workflows/quickstart_ci.yml
@@ -58,10 +58,6 @@ concurrency:
   group: '${{ github.workflow }} @ ${{ github.ref || github.run_id }}'
   cancel-in-progress: true
 
-env:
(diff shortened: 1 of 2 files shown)


jfdenise:
I added this unrelated change: https://github.com/wildfly/quickstart/pull/916/commits/3d174e8b3d9d85ac837849efc96c0f5a917dea3f
Without this fix, ejb-tx-remote-call  quickstart can't succeed.

kabir (approved):
@jfdenise Do you have a bit more info about the -Pprovisioned-server needed for the ejb-txn-remote-call one?
I remember wondering about that some time back

jfdenise:
@kabir , we configure the env there: https://github.com/wildfly/quickstart/blob/main/ejb-txn-remote-call/client/pom.xml#L219 This is done in the profile provisioned-server

PR #946, opened 2024-08-09
[WFLY-19629] set bootableJarName in bootable-jar profile

kstekovi:
https://issues.redhat.com/browse/WFLY-19629

### DIFF
--- a/.github/workflows/quickstart_ci.yml
+++ b/.github/workflows/quickstart_ci.yml
@@ -134,7 +134,7 @@ jobs:
           echo "Building bootable jar..."
           mvn -fae clean ${{ inputs.MVN_COMMAND }} -Pbootable-jar
           echo "Starting bootable jar..."
-          mvn -f ${{ inputs.DEPLOYMENT_DIR }}/pom.xml wildfly:start-jar -Dstartup-timeout=120  ${{ inputs.EXTRA_RUN_ARGS }}
+          mvn -f ${{ inputs.DEPLOYMENT_DIR }}/pom.xml wildfly:start-jar -Dbootable-jar-name=${{ inputs.QUICKSTART_PATH }}-bootable.jar -Dstartup-timeout=120  ${{ inputs.EXTRA_RUN_ARGS }}
           echo "Testing bootable jar..."
           mvn -fae verify -Dserver.host=${{ inputs.SERVER_PROVISIONING_SERVER_HOST }} -Pintegration-testing
           echo "Shutting down bootable jar..."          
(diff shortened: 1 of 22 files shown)


emmartins:
@kstekovi for this kind of solution you need to change the quickstart_ci.yaml lines 137 and 244 to include -Dbootable-jar-name=${{ inputs.QUICKSTART_PATH }}-bootable.jar , this is why all CI is failing on bootable jar testing... and also there is an instruction at https://github.com/wildfly/quickstart/blob/main/shared-doc/run-integration-tests-with-bootable-jar.adoc?plain=1#L22 where you need to do similar

There is an alternative tho, which is to keep the plugin default config i.e. name all bootable jars "server-bootable.jar" (still removing all those not needed <name/> elements tho), and change instead the README instructions at https://github.com/wildfly/quickstart/blob/main/shared-doc/build-and-run-the-quickstart-with-bootable-jar.adoc lines 65 and 68, to use server-bootable.ajr instead of {artifactId}-bootable-jar. Personally I prefer this alternative since it means less pom.xml content, but maybe it is just me.

@jfdenise @jamezp any opinion on what should be the default wildfly maven plugin config we should use wrt bootable jars?

emmartins (changes requested):
See comment above

jamezp:
I need to think about this. The `bootablJarName` really should have been defaulting to `${project.artifactId}-bootable.jar`. It would be an easy thing to change, but isn't backwards compatible. However, it might not really be a big deal either.

jamezp:
FYI https://issues.redhat.com/browse/WFMP-265

emmartins:
@jamezp considering that we didn't yet release anything on EAP yet I agree it shouldn't be a big issue to change to ${project-artifactId}-bootable.jar, and that should mean all QS needs to do is move to the new plugin version

jamezp:
> @jamezp considering that we didn't yet release anything on EAP yet I agree it shouldn't be a big issue to change to ${project-artifactId}-bootable.jar, and that should mean all QS needs to do is move to the new plugin version
> 
> PS: please note that this affects not only package goal, also start-jar

We've had it supported in WildFly since 5.0, which is the current release. It would be a breaking change for 5.0, but might be okay. I'll have to think if that's something we want to do or if it should go into 5.1.

kstekovi:
> @kstekovi for this kind of solution you need to change the quickstart_ci.yaml lines 137 and 244 to include -Dbootable-jar-name=${{ inputs.QUICKSTART_PATH }}-bootable.jar , this is why all CI is failing on bootable jar testing... and also there is an instruction at https://github.com/wildfly/quickstart/blob/main/shared-doc/run-integration-tests-with-bootable-jar.adoc?plain=1#L22 where you need to do similar. You also need to revert the removal of `<name/>` lines, cause those are used to define the web path of the quickstart (from root, being ROOT.war same as "/").
> 
> There is an alternative tho, which is to keep the plugin default config i.e. name all bootable jars "server-bootable.jar", and change instead the README instructions at https://github.com/wildfly/quickstart/blob/main/shared-doc/build-and-run-the-quickstart-with-bootable-jar.adoc lines 65 and 68, to use server-bootable.jar instead of {artifactId}-bootable-jar. Personally I prefer this alternative since it means less pom.xml content, but maybe it is just me.
> 
> @jfdenise @jamezp any opinion on what should be the default wildfly maven plugin config we should use wrt bootable jars?

Hi @emmartins requested change done. I did the first approach even it meas one more line in the pom.xml 

@jamezp regarding to https://issues.redhat.com/browse/WFMP-265 using the `<bootableJarName>` is just a workaround till the WFMP-265 will be done?

emmartins:
@jamezp any idea when we can have WFMP-265 done? Should we proceed with this one till the plugin is updated?

kstekovi:
@jamezp does is WFMP-265 only about a default value? So with the `-Dbootable-jar-name=microprofile-config-bootable.jar` parameter it should look for `microprofile-config-bootable.jar`?

emmartins:
@kstekovi maybe that is a second issue, @jamezp can you please look at that one too?

emmartins:
no longer needed since the plugin is now using to the old wildfly-jar behavior, producing ${artifactId}-bootable.jar

jamezp:
Oh wow. I don't know how I missed this. I'm sorry about that @emmartins. I think it's time to clean up my inbox :)

PR #948, opened 2024-08-26
[WFLY-19662] Switch from / to helloworld-ws context in QS

baranowb:
https://issues.redhat.com/browse/WFLY-19662

### DIFF
--- a/helloworld-ws/README.adoc
+++ b/helloworld-ws/README.adoc
@@ -503,7 +503,7 @@ Review the server log to see useful information about the deployed web service e
 [source,options="nowrap"]
 ----
 JBWS024061: Adding service endpoint metadata: id=org.jboss.as.quickstarts.wshelloworld.HelloWorldServiceImpl
- address=http://localhost:8080/HelloWorldService
+ address=http://localhost:8080/helloworld-ws/HelloWorldService
  implementor=org.jboss.as.quickstarts.wshelloworld.HelloWorldServiceImpl
  serviceName={http://www.jboss.org/eap/quickstarts/wshelloworld/HelloWorld}HelloWorldService
  portName={http://www.jboss.org/eap/quickstarts/wshelloworld/HelloWorld}HelloWorld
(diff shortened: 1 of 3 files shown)


emmartins:
Unfortunately the changes don't work with server provisioning and openshift, thus the test fails, and after some time investigating I found no way to easily support the current dual mode we have in other quickstarts, that this issue wants to implement (default deploy at /helloworld-ws for baremetal, deploy at root for openshift). I mean you either deploy always at root (current design), or you always deploy at /helloworld-ws, you can't have it working for both cases depending on wether using baremetal or not.

In theory the only changes that would be needed in this PR, besides test logic, would be to delete the jboss-web.xml, letting undertow decide whether to deploy at root or not, but if you do this then deploy at root doesn't work (this is an old issue, see https://developer.jboss.org/thread/248731). @jimma maybe you want to investigate this particular issue (why removing the jboss-web.xml no longer works for "mvn package -Pprovisioned-server" + "mvn wildfly:start") in a separate WFLY jira?

Anyway, please note that there is another JIRA that plans to do kind of related modifications, https://issues.redhat.com/browse/WFLY-19463 , more specifically the "No deployment at web root" change, which will do similar changes to this PR but also removing the ROOT.war config in the provisioned-server maven profile, and then it works for both baremetal and provisioning-server (full changes for that at https://github.com/emmartins/quickstart/pull/new/WFLY-19662 ) so I think let's just defer this fix to that JIRA on WFLY 35. WDYT @baranowb @kstekovi ?

jimma:
@emmartins  I already created https://issues.redhat.com/browse/WFLY-19780. I'll try to look more things to see if this can be fixed without breaking other things. I'll keep you updated once I find this can't be fixed without breaking other things.

emmartins:
thank you @jimma

baranowb:
I could have sworn Ive replied to this already. I have no preference, just something I noticed when I was working on some issue, so I filled PRs.

kstekovi:
Hi @emmartins 

The https://issues.redhat.com/browse/WFLY-19463 is resolved. can we finish this one?

jimma:
@emmartins https://github.com/wildfly/wildfly/pull/18290 is still not merged. Without this change, it seems this doesn't work ?

emmartins:
It works due to additional changes merged after this one, we now never
deploy the quickstart at root so we don’t hit the issue being fixed by the
wildfly PR.

PR #953, opened 2024-09-16
[WFLY-19618] Helloworld REST Quickstart

theashiot:
Issue: https://issues.redhat.com/browse/WFLY-19618


emmartins (commented):
Let's change a bit the Quickstart, to become the helloworld with only one change, we replace the servlet class with the classes for the REST app and the REST endpoint, in short considering this PR we do the following changes:

- remove the cdi bean HelloService
- we sync the test class with helloworld quickstart (you may improve both, but we should have both in sync)
- we change the HelloWorld REST endpoint to interface with the user exactly as helloworld:
```
/*
 */
package org.jboss.as.quickstarts.rshelloworld;

import jakarta.inject.Inject;
import jakarta.ws.rs.GET;
import jakarta.ws.rs.Path;
import jakarta.ws.rs.Produces;

/**
 * A simple REST service which is able to say "Hello World!"" at /HelloWorld
 *
 */
@Path("/")
public class HelloWorld {
    @GET
    @Path("/HelloWorld")
	@Produces(MediaType.TEXT_PLAIN)
    public String getHelloWorldJSON() {
        return "<h1>Hello World!</h1>";
    }
} 
```
- update the README-source according to changes above

helloworld-rs/README-source.adoc
```diff
@@ -0,0 +1,57 @@
+include::../shared-doc/attributes.adoc[]
+
+= helloworld-rs: Jakarta RESTful Web Services Helloworld Example
+:author: Ashwin Mehendale
+:level: Beginner
+:technologies: Jakarta RESTful Web Services
+:openshift: true
+:portedToGlow: false
```
emmartins, line 8: in case you don't overwrite this with the helloworld README, please remember to move this to true

theashiot: done

helloworld-rs/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+            <releases>
+                <enabled>true</enabled>
+                <updatePolicy>never</updatePolicy>
+            </releases>
+            <snapshots>
+                <enabled>true</enabled>
+                <updatePolicy>never</updatePolicy>
+            </snapshots>
+            <layout>default</layout>
+        </repository>
```
jamezp, line 69: There should be no need for this repository.

helloworld-rs/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+            <id>redhat-ga-maven-repository</id>
+            <name>Red Hat GA Maven Repository</name>
+            <url>https://maven.repository.redhat.com/ga/</url>
+            <releases>
+                <enabled>true</enabled>
+            </releases>
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+        </pluginRepository>
```
jamezp, line 93: There should be no need for this repository.

helloworld-rs/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+                <artifactId>wildfly-ee-with-tools</artifactId>
+                <version>${version.bom.ee}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <!-- Import the JAX-RS API, we use provided scope as the API is included in JBoss EAP -->
```
jamezp, line 110: We should replace "JAX-RS" with Jakarta REST. Also, should we reference EAP here?

helloworld-rs/src/main/java/org/jboss/as/quickstarts/helloworldrs/HelloWorld.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+@Path("/")
+public class HelloWorld {
+
+    @GET
+    @Path("/HelloWorld")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String getHelloWorldJSON() {
+        return "<h1>Hello World!</h1>";
```
jamezp, line 25: The HTML tags should be removed for plain text probably.

helloworld-rs/src/main/java/org/jboss/as/quickstarts/helloworldrs/JakartaRESTActivator.java
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.jboss.as.quickstarts.rshelloworld;
+
+import jakarta.ws.rs.ApplicationPath;
+import jakarta.ws.rs.core.Application;
+
+/**
+ * JakartaRESTActivator is an arbitrary name, what is important is that jakarta.ws.rs.core.Application is extended
+ * and the @ApplicationPath annotation is used with a "rest" path.  Without this the rest routes linked to
```
jamezp, line 12: Minor, but instead of "rest routes" I think we should use "rest endpoints".

helloworld-rs/src/test/java/org/jboss/as/quickstarts/helloworldrs/BasicRuntimeIT.java
```diff
@@ -0,0 +1,70 @@
...
+                .uri(new URI(serverHost+"/rest/HelloWorld"))
+                .GET()
+                .build();
+        final HttpClient client = HttpClient.newBuilder()
+                .followRedirects(HttpClient.Redirect.ALWAYS)
+                .connectTimeout(Duration.ofMinutes(1))
+                .build();
+        final HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+        assertTrue(response.body().contains("Hello World!"));
+    }
```
jamezp, line 67: It seems a little strange to me to have two different tests here. Both should likely be asserting on the status code.

Also, is there a reason not to use the REST Client given this is a REST quickstart?

helloworld-rs/src/main/java/org/jboss/as/quickstarts/helloworldrs/HelloWorld.java
```diff
@@ -0,0 +1,27 @@
...
+ * @author emartins
+ */
+
+@Path("/")
+public class HelloWorld {
+
+    @GET
+    @Path("/HelloWorld")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String getHelloWorldJSON() {
```
jamezp, line 24: The method name should be changed as this isn't a JSON response.

Also should we add a `@PathParam("name")` and great with the name instead of world? e.g. 

```java
@GET
@Path("/hello/{name}")
@Produces(MediaType.TEXT_PLAIN)
public String hello(@PathParam("name") final String name) {
    return String.format("Hello %s!", name);
}
```

Not needed, but just shows a simple path parameter too.

jamezp (changes requested):
There should probably be a CI workflow for this as well.

helloworld-rs/src/main/java/org/jboss/as/quickstarts/helloworldrs/HelloWorld.java
theashiot: done

helloworld-rs/src/main/java/org/jboss/as/quickstarts/helloworldrs/JakartaRESTActivator.java
theashiot: done

theashiot:
Thanks, @jamezp for the review! I've started implementing the changes. I'll let you know when the PR is ready for review again.

cheers,
ashwin

helloworld-rs/pom.xml
emmartins: +1

helloworld-rs/pom.xml
emmartins: +1

emmartins: This change should extend to the guide itself, by default users should only add the jboss.org Repo, but we should give optional instructions wrt adding also GA repo, since some apps may require it (e.g. microprofile-jwt quickstart).

helloworld-rs/pom.xml
emmartins: +1

emmartins: This seems to affect other Quickstarts, so I opened https://issues.redhat.com/browse/WFLY-19804 to dela with this.

helloworld-rs/src/test/java/org/jboss/as/quickstarts/helloworldrs/BasicRuntimeIT.java
emmartins: +1

helloworld-rs/src/main/java/org/jboss/as/quickstarts/helloworldrs/HelloWorld.java
emmartins: +1

helloworld-rs/pom.xml
theashiot: removed

helloworld-rs/pom.xml
theashiot: removed

helloworld-rs/pom.xml
theashiot: Updated to: " Import the Jakarta REST API, we use provided scope as the API is included in the server"

helloworld-rs/src/test/java/org/jboss/as/quickstarts/helloworldrs/BasicRuntimeIT.java
theashiot: Created a new class that uses a REST client: https://github.com/wildfly/quickstart/pull/953/files#diff-76464e0c59b997d4e308fd7c0f653e92733f182358fe8ef6f620f132347e10cd

helloworld-rs/src/main/java/org/jboss/as/quickstarts/helloworldrs/HelloWorld.java
theashiot: updated the method name.

theashiot:
Thanks a lot @jamezp and @emmartins for the review! I've made the changes. Can you please have another look?

best,
ashwin

helloworld-rs/pom.xml
```diff
@@ -0,0 +1,225 @@
...
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>commons-logging-jboss-logging</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-jackson2-provider</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jamezp, line 97: We should be using `org.jboss.resteasy:resteasy-json-p-provider` here. Jackson will work fine, but the Jakarta JSON Processing API is our default and preferred provider.

jamezp (changes requested):

helloworld-rs/pom.xml
emmartins: Please add the following dependency
```
        <!-- Required by the resteasy-json-p-provider -->
        <dependency>
            <groupId>org.eclipse.parsson</groupId>
            <artifactId>parsson</artifactId>
            <scope>test</scope>
        </dependency>
```

theashiot: Thanks, @emmartins, that fixed it!

jamezp: Ah, this looks like a bug in the RESTEasy dependency. The dependency is marked as optional, but it really shouldn't be. I've filed https://issues.redhat.com/browse/RESTEASY-3548 to resolve that.

emmartins (approved):

jamezp (approved):

PR #956, opened 2024-09-24
[WFLY-19772] Optimizes Maven Repositories configurations

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-19772

Change log:
- Removes Red Hat Repositories from all QS that do not require it (i.e. all except microprofile-jwt)
- Moves Repositories configuration to bottom of the Maven Project, so all match same code style as QS lead by James
- Fixes microprofile-config Maven Project indent (to match all other QS)

### DIFF
--- a/batch-processing/pom.xml
+++ b/batch-processing/pom.xml
@@ -51,62 +51,6 @@
         <version.plugin.wildfly>5.0.0.Final</version.plugin.wildfly>
     </properties>
 
-
-    <repositories>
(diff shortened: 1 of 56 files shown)


emmartins:
Please note that we are expecting fails for "WITH DEPS" on Project CI and all MicroProfile BOM dependent quickstarts, due to the BOM rename merged on WFLY main branch.

PR #960, opened 2024-09-30
[WFLY-19790] Remove batch-processing DS file

liweinan:
Issue: https://issues.redhat.com/browse/WFLY-19790


liweinan:
@jfdenise I don't know how to fix this error:

- https://github.com/wildfly/quickstart/actions/runs/11109598030/job/30865320646?pr=960#step:6:6605

As the `h2-database` has already been added. Could you please help check this? Thanks!

emmartins (changes requested):

emmartins:
Besides the script change, guess we need instructions wrt the script execution when used with bare-metal standard distribution, and a config restore script too? You can reuse content from https://github.com/wildfly/quickstart/blob/484ba35272679919fca411d94e0072f0513c36b6/ee-security/README-source.adoc?plain=1#L40

emmartins:
Btw I really thought you would just need to update the persistence.xml, ie no need for config change script...
@scottmarlow I wonder if this (replace usage of the DS file with just persistence.xml) would be possible?

liweinan:
I discussed this issue with @jamezp, and the DS file may still be used in the future, so I am holding off on this work.

batch-processing/configure-server.cli
```diff
@@ -0,0 +1,15 @@
...
+# Start by creating the JDBC datasource
+/subsystem=datasources/data-source=batch-processingDS:add(connection-url="jdbc:h2:mem:batch-processingDS;DB_CLOSE_ON_EXIT=FALSE;DB_CLOSE_DELAY=-1", jndi-name="java:jboss/datasources/batch-processingDS", driver-name=h2, user-name="sa", password="sa", enabled="true", use-java-context=true)
+/subsystem=batch-jberet/jdbc-job-repository=jdbc:add(data-source=batch-processingDS)
+/subsystem=batch-jberet:write-attribute(name=default-job-repository,value=jdbc)
+
+# Run the batch commands
+run-batch
+
+# Reload the server configuration
+#reload
```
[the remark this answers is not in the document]
liweinan: I can create another PR to do this if necessary.

emmartins:
@jamezp can you please clarify what @liweinan said above about new use for DS files? We have those deprecated for long, and we are logging a log warning (the reason for this issue) when deploying such files...

liweinan:
> One option is to simply use the default data source

btw here is one blog post that I'm writing on this topic: https://github.com/wildfly/wildfly.org/pull/677

If this is a better solution for this example, I can work on it.

emmartins:
really up to you guys what you want to use as datasource, be it standard or not I'm fine with it, just agree with QE it's no longer ok to use a -ds file to config a database, you now have the standard persistence.xml and DatasourceDefinition annotation that should be the alternative, or at least the first alternative, if the vendor requires something more (like a cli script then be it)

liweinan:
James has provided an example showing the CLI usage here: https://github.com/jberet/jberet-examples/pull/13

I'll use

liweinan:
@emmartins I have replaced Glow with the data source feature pack in combination with the CLI script to create the database and configure the datasource. Could you please check if this solution is acceptable? Thanks!

liweinan:
@jamezp I'm not sure how to fix the Kubernetes CI test failure. Could you please help to check it when you have time? Thanks!

jamezp:
@liweinan I have no idea why the Kubernetes tests are failing. There seems to be no indication as to why.

batch-processing/pom.xml
```diff
@@ -200,12 +203,32 @@
                         <groupId>org.wildfly.plugins</groupId>
                         <artifactId>wildfly-maven-plugin</artifactId>
                         <configuration>
-                            <discover-provisioning-info>
```
emmartins, line 203: why are we going back to directly using galleon layers, instead of glow's discovery? This seems a mistake

emmartins: Maybe you are just doing copy paste of outdated (in the sense of not using glow) configuration? You can use glow and pack a script as shown at https://github.com/wildfly/quickstart/blob/82ceea0a283cd1b7a2d1bdfd51a8c41b23e8b182/opentelemetry-tracing/pom.xml#L161

batch-processing/add-h2-ds.cli
```diff
@@ -0,0 +1 @@
+xa-data-source add --name=batch_db --enabled=true --use-java-context=true --use-ccm=true --jndi-name=java:jboss/datasources/batch-processingDS --xa-datasource-properties={"URL"=>"jdbc:h2:mem:batch-processing;DB_CLOSE_ON_EXIT=FALSE;DB_CLOSE_DELAY=-1"} --driver-name=h2 --password=sa --user-name=sa --same-rm-override=false --no-recovery=true
```
emmartins, line 1: I still think you could do this op using only Jakarta specs, which would make this app portable among vendors. But I tried to use a DataSourceDefinition like the one below (you can add this for instance to Contact class) but boot failed with a CNFE for org.h2.jdbcx.JdbcDataSource. I guess we can try to explore the specs way and clarify how should users use the specs DataSourceDefinition later

@DataSourceDefinition(name="java:jboss/datasources/batch-processingDS",
        className="org.h2.jdbcx.JdbcDataSource",
        url="jdbc:h2:mem:batch-processing;DB_CLOSE_ON_EXIT=FALSE;DB_CLOSE_DELAY=-1",
        user="sa",
        password="sa"
)

batch-processing/pom.xml
```diff
@@ -231,13 +254,32 @@
...
-                                <version>${version.server}</version>
-                                <context>cloud</context>
-                                <addOns>
-                                    <addOn>h2-database:default</addOn>
-                                </addOns>
-                            </discover-provisioning-info>
+                            <feature-packs>
```
emmartins, line 257: we should revert this change too, other than adding the script of course

emmartins (changes requested):
Please see my comments

batch-processing/add-h2-ds.cli
emmartins: nevermind, I was able to make it work, I will send an alternative PR for you to have a look

emmartins:
@liweinan @jamezp alternative PR using the specs DataSourceDefinition at https://github.com/wildfly/quickstart/pull/973

liweinan:
@emmartins Thanks for the review! I'll check it.

> alternative PR using the specs DataSourceDefinition at https://github.com/wildfly/quickstart/pull/973

Cool! I'll see if this can work with this example.

liweinan:
I closed this PR because it's been replaced by the PR mentioned above.

PR #973, opened 2024-10-30
[WFLY-19790] batch-processing: replaces -ds.xml deprecated filed with Jakarta's DataSou…

emmartins:
…rceDefinition
Issue: https://issues.redhat.com/browse/WFLY-20905


emmartins:
@tadamski I am seeing the following warning when using the DataSourceDefinition instead of the -ds.xml

20:12:46,089 WARN  [org.jboss.jca.core.connectionmanager.pool.strategy.OnePool] (ServerService Thread Pool -- 46) IJ000407: No lazy enlistment available for batch-processing_batch-processing_batch-processing_java:jboss/datasources/batch-processingDS

Perhaps there is a property we need to add to the DataSourceDefinition to prevent this?

emmartins:
@jfdenise FYI it seems Glow is not aware of DataSourceDefinition annotation and reports the following error in this project, if you remove the failOnError

`[ERROR] * unbound datasources error: java:jboss/datasources/batch-processingDS`

emmartins:
@liweinan IMHO this is still a work in progress, after all it currently replaces one warn with another, and then there is the provisioning error... I suspect we will need server side and glow side changes to make it "proper"!

@kstekovi WDYT?

liweinan:
@emmartins cool!

emmartins:
@tadamski can you please have a look at my comment above once you have a chance?

emmartins:
@kstekovi for the record I don't think we should merge this till we have that new warning suppressed or documented, we should not replace one warning with another.

jfdenise:
@emmartins @liweinan , 
DataSourceDefinition is going to be natively supported by WildFly Glow. We have the following PR opened:
* https://github.com/wildfly/wildfly-glow/pull/94
* https://github.com/wildfly/wildfly/pull/18420
Once merged this would require a WildFly Maven Plugin upgrade and usage of WildFly SNAPSHOT (until WF 35.Beta1 is released).

You would then have to remove the explicit h2-datasource add-on, and the fail-on-error.

jfdenise:
@emmartins , FYI, The changes in WildFly have been merged. We are waiting for a new WildFly Maven Plugin release that integrates the support.

emmartins:
@tadamski please please? ^^

tadamski (approved):

emmartins:
@tadamski thanks for your feedback, personally I see no interest in making this change till the warning is gone, after all this would be replacing one warning (that is already documented) with another.

tadamski:
@emmartins This warning is only informative, tbh I was considering removing it, but after it was downgraded to debug it won't be visible anyway in standard configuration

emmartins:
@tadamski does this means that WFLY 36 will no longer produce this warning?

tadamski:
@emmartins yes

PR #985, opened 2024-11-28
[WFLY-20032] disable tls on OpenShift for Thread racing QS

kstekovi:
https://issues.redhat.com/browse/WFLY-20032


emmartins:
Supersed by #986

PR #986, opened 2024-12-03
[WFLY-20032] adds sslcontext, which trusts any cert, to thread-racing…

emmartins:
… jaxrs stage
Issue: https://issues.redhat.com/browse/WFLY-20032

### DIFF
--- a/thread-racing/pom.xml
+++ b/thread-racing/pom.xml
@@ -150,7 +150,7 @@
         </dependency>
         <dependency>
             <groupId>org.jboss.resteasy</groupId>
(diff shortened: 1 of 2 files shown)


emmartins:
@kstekovi here is a solution for your issue, you may try it by pointing your local helm chart to 

```
  uri: https://github.com/emmartins/quickstart.git
  ref: thread-racing-pkix
```

I am still not sure this is something we should do in the app, I would prefer a proper setup of OpenShift since that is what users should do, but I dunno what is the alternative...

PR #987, opened 2024-12-04
[WFLY-19913] Requested modifications have been implemented

jmfinelli:
[WFLY-19913](https://issues.redhat.com/browse/WFLY-19913)


emmartins:
@jmfinelli so far it looks good, just need to figure out what we will do wrt to the "always on server provisioning" issue

emmartins (approved):

PR #990, opened 2024-12-13
[WFLY-19337] The GitHub Action to test ejb-txn-remote-call is now working

jmfinelli:
[WFLY-19337](https://issues.redhat.com/browse/WFLY-19337)


kabir (changes requested):
I'm guessing the tests are failing due to the refactoring of the script.

Run some of the other tests and adjust as needed. Once it all passes I can review properly :-)

@emmartins please comment to say whether you agree/disagree with my below comments

.github/workflows/scripts/kubernetes/core/run-quickstart-test-on-kubernetes.sh
```diff
@@ -130,35 +104,15 @@ runPostHelmInstallCommands
 echo "running the tests"
 pwd
 
-route="localhost:8080"
```
kabir, line 133: I"m not sure I like 'hiding' this in the overridable-functions.

I think it would be cleaner to have some kind of check that can be overridden per tests in overridable-functions.sh, and an if/else check here. In the normal (if) case, run the tests normally in this file, and if we should run the tests ourselves implement that in the tests overridable-functions.sh.

BTW, I think a few other quickstarts use client/server, so if that can be a common use case which can be handled here, even better. Although perhaps that can come later.

.github/workflows/scripts/kubernetes/core/run-quickstart-test-on-kubernetes.sh
```diff
@@ -76,49 +76,23 @@ fi
...
-
-echo "Tagging image and pushing to registry..."
-export root_image_name="localhost:5000/${application}"
-export image="${root_image_name}:latest"
-docker tag ${qs_dir} ${image}
-docker push ${image}
+provisionServer "${application}" "${qs_dir}"
 
 ################################################################################################
-# Helm install, waiting for the pods to come up
```
kabir, line 95: Same observation as for the running of the tests. IMO the default behaviour should remain in this file, and a check should be done for when we don't want to do that, and handle that accordingly.

.github/workflows/scripts/kubernetes/core/run-quickstart-test-on-kubernetes.sh
```diff
@@ -130,35 +104,15 @@ runPostHelmInstallCommands
...
-mvn -B verify -Pintegration-testing ${mvnVerifyArguments}
-
-if [ "$?" != "0" ]; then
-  test_status=1
-  echo "Tests failed!"
-  echo "Dumping the application pod(s)"
-  kubectl logs deployment/"${application}"
-  testsFailed
-fi
+runningTests "${application}" "${server_protocol}" "$(getMvnVerifyExtraArguments)"
```
kabir, line 107: I"m not sure I like 'hiding' this in the overridable-functions.

I think it would be cleaner to have some kind of check that can be overridden per tests in overridable-functions.sh, and an if/else check here. In the normal (if) case, run the tests normally in this file, and if we should run the tests ourselves implement that in the tests overridable-functions.sh.

BTW, I think a few other quickstarts use client/server, so if that can be a common use case which can be handled here, even better. Although perhaps that can come later.

.github/workflows/scripts/kubernetes/core/run-quickstart-test-on-kubernetes.sh
```diff
@@ -76,49 +76,23 @@ fi
...
 
-echo "Building application and provisioning server image..."
-mvn -B package -Popenshift wildfly:image -DskipTests
-
-echo "Tagging image and pushing to registry..."
-export root_image_name="localhost:5000/${application}"
-export image="${root_image_name}:latest"
-docker tag ${qs_dir} ${image}
-docker push ${image}
+provisionServer "${application}" "${qs_dir}"
```
kabir, line 86: Same comment. Please keep the default behavious here, and if something special is needed do that in the test's overridable-functions

emmartins:
@kabir if I understood correctly the issue with this proposal is that if a QS wants to override a function, it needs to override all, and you are proposing an alternative where that doesn't need to happen?

kabir:
@emmartins More or less, I'd like to keep the code that works in all other quickstarts apart from this one in the main script. Something like:
```
# 'unstandard_check' is from the main overrides, and can be overridded per test if it needs something else 
if [ $unstandard_check == 0]; then
   //Current bash content
else
  # From the test overrides
  $(unstandard_impl $x $y $z)
fi
```
Rather than putting '//Old bash content' into the mai

ejb-txn-remote-call/README-source.adoc
```diff
@@ -293,7 +296,7 @@ the `client` and `server` application can be deployed
 [source,sh,subs="+quotes,attributes+",options="nowrap"]
 ----
 cd ${PATH_TO_QUICKSTART_DIR}/
-mvn clean package
+mvn clean package -P '!provisioned-server'
```
emmartins, line 299: leftover?

ejb-txn-remote-call/server/pom.xml
```diff
@@ -207,70 +209,38 @@
...
-                                    </discover-provisioning-info>
-                                </configuration>
-                            </execution>
-                        </executions>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-        <profile>
-            <id>openshift</id>
```
emmartins, line 238: we no longer have openshift profile in server module, is this expected?

jmfinelli: Yep, I realised we can just trust the `provisioned-server` profile

.github/workflows/scripts/kubernetes/core/run-quickstart-test-on-kubernetes.sh
jmfinelli: Thanks for your review, @kabir. I hope the last commit is close enough to having a default behaviour. Please, let me know what you think about it :-)

.github/workflows/scripts/kubernetes/core/run-quickstart-test-on-kubernetes.sh
jmfinelli: Thanks for your review, @kabir. I hope the last commit is close enough to having a default behaviour. Please, let me know what you think about it :-)

.github/workflows/scripts/kubernetes/core/run-quickstart-test-on-kubernetes.sh
jmfinelli: Thanks for your review, @kabir. I hope the last commit is close enough to having a default behaviour. Please, let me know what you think about it :-)

.github/workflows/scripts/kubernetes/core/overridable-functions.sh
```diff
@@ -43,12 +46,53 @@ function installPrerequisites()
...
+  echo "Nothing to do in provisionServer()..."
+}
+
+# Trigger a custom behaviour when it comes to
+# setting up the environment
+# Returns
+# 0 - false
+# 1 - true
+#
+function customProcessing() {
```
kabir, line 84: 'processing' feels a bit vague, and 'helmInstall' is already used. Maybe 'customHelmProcessing'?

kabir (commented):
Aside from needing some conflicts resolved, and a comment about one of the names, I think this now looks more like what I had in mind :-)

.github/workflows/scripts/kubernetes/core/overridable-functions.sh
jmfinelli: At least for `ejb-tx-remote-call`, this function is not about Helm. That's why I went for the generic `processing`. I'm very happy to go for something else (as I don't like `processing` myself) but I'd like something without helm in the function's name. WDYT?

kabir: Hmm, maybe something like 'customDeploy'? I think this has the logic of how we deploy the app, and covers both the Helm and non-Helm cases?

jmfinelli: Cool. `deploy` and `customDeploy` will be then :-) Thanks Kabir

kabir (approved):

PR #998, opened 2024-12-20
[WFLY-20210] Removed '/' from the qs_dir variable

jmfinelli:
[WFLY-20210](https://issues.redhat.com/browse/WFLY-20210)


emmartins:
ty @jmfinelli

PR #1007, opened 2025-01-21
[WFLY-19725] Document jaxrs-jwt script execution warnings

emmartins:
Issue: https://issues.redhat.com/browse/WFLY-19725


jaxrs-jwt/README-source.adoc
```diff
@@ -61,6 +61,19 @@ $ __{jbossHomeName}__/bin/jboss-cli.sh --connect --file=configure-elytron.cli
...
+[source]
+----
+WFLYELY01090: Allowed jku values haven't been specified for token realm 'jwt-realm'. Token validation will fail if the token contains a 'jku' header parameter. The allowed jku values can be specified as a space separated string using the 'wildfly.elytron.jwt.allowed.jku.values.jwt-realm' system property.
+
+ELY01179: SSL not configured. jku claim will not be supported.
+
+ELY01182: Allowed jku values haven't been configured for the JWT validator. Token validation will fail if the token contains a 'jku' header parameter.
+----
++
+The warnings are due to jku token validation not being used, to simplify the quickstart.
```
fjuma, line 75: s/The warnings are due to jku token validation not being used, to simplify the quickstart./The quickstart creates and makes use of a JWT that doesn't include a "jku" header parameter so the warnings can be ignored. When making use of a real token provider with a `jwt-realm`, token validation will fail if the token contains a "jku" header parameter and the allowed jku values have not been specified.

fjuma (changes requested):

jaxrs-jwt/README-source.adoc
emmartins: thank you

fjuma (approved):
Thanks @emmartins.

PR #1009, opened 2025-01-23
[WFLY-20129] Mail QS tests - Use HtmlUnit instead of Selenium

kstekovi:
https://issues.redhat.com/browse/WFLY-20129

### DIFF
--- a/mail/pom.xml
+++ b/mail/pom.xml
@@ -51,9 +51,8 @@
         <!-- the versions for BOMs, Packs and Plugins -->
         <version.bom.ee>${version.server}</version.bom.ee>
         <version.plugin.wildfly>5.1.3.Final</version.plugin.wildfly>
-        <!-- the versions for other Dependencies and Plugins -->
-        <version.org.seleniumhq.selenium>4.15.0</version.org.seleniumhq.selenium>
-        <version.webdrivermanager>5.7.0</version.webdrivermanager>
+        <!-- the version for HtmlUnit -->
+        <version.org.htmlunit>4.15.0</version.org.htmlunit>
     </properties>
 
     <dependencyManagement>
@@ -100,16 +99,9 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.seleniumhq.selenium</groupId>
-            <artifactId>selenium-java</artifactId>
-            <version>${version.org.seleniumhq.selenium}</version>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>io.github.bonigarcia</groupId>
-            <artifactId>webdrivermanager</artifactId>
-            <version>${version.webdrivermanager}</version>
+            <groupId>org.htmlunit</groupId>
+            <artifactId>htmlunit</artifactId>
+            <version>${version.org.htmlunit}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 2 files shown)


emmartins:
@Dkafetzis FYI

mail/pom.xml
```diff
@@ -100,17 +100,15 @@
...
+            <artifactId>htmlunit</artifactId>
+            <version>${version.org.htmlunit}</version>
             <scope>test</scope>
         </dependency>
-
         <dependency>
-            <groupId>io.github.bonigarcia</groupId>
-            <artifactId>webdrivermanager</artifactId>
-            <version>${version.webdrivermanager}</version>
-            <scope>test</scope>
```
emmartins, line 113: the test scope should stay

mail/pom.xml
```diff
@@ -100,17 +100,15 @@
...
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>${version.org.apache.commons}</version>
```
emmartins, line 111: This artifact is in the ee BOM so you don't need to manage this version.

mail/src/test/java/org/jboss/as/quickstarts/mail/MailTestCaseIT.java
```diff
@@ -1,145 +1,97 @@
...
-        body.sendKeys("Hello user02, I've sent an email.");
-
-        WebElement submitButton = driver.findElement(By.id("smtp_send_btn"));
         submitButton.click();
+        /* will wait JavaScript to execute up to 30s */
+        webClient.waitForBackgroundJavaScript(30 * 1000);
 
-        wait.until(d -> driver.findElement(By.xpath("//ul[@id='smtp_messages']/li")).isDisplayed());
+        HtmlElement message = mailHomePage.getFirstByXPath("//ul[@id='smtp_messages']/li");
+        Assert.assertEquals("Unexpected result messages after sending an email via SMTP.", "Email sent to user02@mail.local", message.asNormalizedText());
```
emmartins, line 65: please remove this empty line

emmartins (changes requested):
Sorry, noticed a few minor issues.

mail/pom.xml
kstekovi: Unfortunately the `wildfly-ee` bom contains this artifact in version `3.17.0` which is not compatible with htmlunit. 

With the version `3.17.0` the test ends with following 

> [ERROR]   MailTestCaseIT.testSendAndRetrieveEmail:36 » NoClassDefFound org/apache/commons/lang3/Strings

I found the `commons-lang3` in version `3.17.0` contains a vulnerability

> Uncontrolled Recursion vulnerability in Apache Commons Lang. This issue affects Apache Commons Lang: Starting with commons-lang:commons-lang 2.0 to 2.6, and, from org.apache.commons:commons-lang3 3.0 before 3.18.0. The methods ClassUtils.getClass(...) can throw StackOverflowError on very long inputs. Because an Error is usually not handled by applications and libraries, a StackOverflowError could cause an application to stop. Users are recommended to upgrade to version 3.18.0, which fixes the issue. Mend Note: The description of this vulnerability differs from MITRE.

Do we can update version of the `commons-lang` in the `wildfly-ee` BOM?

emmartins:
The fail is due to wildfly/wildfly main currently using outdated WildFly Core 29.0.0.Final (which uses commons 3.17.0), there is already a PR open to bump this version at https://github.com/wildfly/wildfly/pull/19164 , which will fix this test issue.

emmartins:
good stuff, thank you @kstekovi 👍

PR #1024, opened 2025-03-05
[WFLY-20447] Fix mistakes in "http custom mechanism" readme

mskacelik:
Issue: https://issues.redhat.com/browse/WFLY-20447

While going through this quickstart I probably found some mistakes in the readme.
- wrong path (.../modules/modules/...)
- one section is copy pasted and is twice in the text
- no mention of the restoring scripts

Also bit off topic, but how exactly does the password validation work for the custom mechanism, given that `X-PASSWORD` could be anything and response is still OK (200) (unlike the `X-USERNAME`)

### DIFF
--- a/http-custom-mechanism/README-source.adoc
+++ b/http-custom-mechanism/README-source.adoc
@@ -105,7 +105,7 @@ $ __{jbossHomeName}__/bin/jboss-cli.sh --connect --file=add-custom-module.cli
 
 NOTE: For Windows, use the `__{jbossHomeName}__\bin\jboss-cli.bat` script.
 
-This creates the custom `__{jbossHomeName}__/modules/modules/org/jboss/as/quickstart/http_custom_mechanism/main` folder, then copies in the `custom-module/target/custom-module.jar` file and creates the required `module.xml` file.
+This creates the custom `__{jbossHomeName}__/modules/org/jboss/as/quickstart/http_custom_mechanism/main` folder, then copies in the `custom-module/target/custom-module.jar` file and creates the required `module.xml` file.
 
 You can verify the module structure in your file manager.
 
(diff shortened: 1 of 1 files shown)


emmartins (changes requested):
hi @mskacelik, thanks for this. I have a few changes to request, you should only change README-source.adoc, the README.adoc is just a flattened version of README-source.adoc that is auto updated after each PR merge. 

So the other README.adoc changes you did should actually be around the following include, that is wrongly done twice:

```
:restoreScriptName: restore-configuration.cli
include::../shared-doc/restore-standalone-server-configuration-manual.adoc[leveloffset=+2]
```

The problem here is that such "included" content does not really supports multiple scripts, so now you may just remove the second include (line 238), and I will create a new JIRA to add multiple script support to that shared docs content; alternatively you can do that extra too in this PR.

emmartins:
@mskacelik by the way, do you have already a WildFly JIRA account? I created a JIRA for your contribution at https://issues.redhat.com/browse/WFLY-20447 , if you have an account I may assign it to you, otherwise I can help as proxy, no prob.

emmartins:
@fjuma it's a bit unusual place for user requests, but perhaps you want to reply here to @mskacelik ?

mskacelik:
> hi @mskacelik, thanks for this. I have a few changes to request, you should only change README-source.adoc, the README.adoc is just a flattened version of README-source.adoc that is auto updated after each PR merge.
> 
> So the other README.adoc changes you did should actually be around the following include, that is wrongly done twice:
> 
> ```
> :restoreScriptName: restore-configuration.cli
> include::../shared-doc/restore-standalone-server-configuration-manual.adoc[leveloffset=+2]
> ```
> 
> The problem here is that such "included" content does not really supports multiple scripts, so now you may just remove the second include (line 238), and I will create a new JIRA to add multiple script support to that shared docs content; alternatively you can do that extra too in this PR, by improving the /shared-doc/restore-standalone-server-configuration-manual.adoc now.

Shouldn't the included file on line 226 be: `restore-standalone-server-configuration.adoc` instead of `restore-standalone-server-configuration-manual.adoc`? This explains the `:restoreScriptName: restore-configuration.cli` above it as a parameter (in the manual, restoreScriptName is not used) and also why the manual is included below it (since it continues the non-manual section nicely).

emmartins (approved):

emmartins:
@mskacelik  thank you :-)

PR #1027, opened 2025-03-17
[WFLY-20424] ejb-multi-server: specify infinispan marshaller configur…

tadamski:
…ation to avoid LEGACY one being used


emmartins:
thank you, will review

emmartins:
warn is gone, thanks @tadamski

PR #1035, opened 2025-05-14
Bump linux version in microprofile-reactive-messaging-kafka CI job, t…

emmartins:
…o ubuntu-latest


emmartins:
@kabir I am changing this since the old version is now deprecated and it will not longer be supported by GH

PR #1038, opened 2025-06-03
[WFLY-20678] ejb-txn-remote-call QS logs waring JGRP000015: the receive buffer of socket MulticastSocket was set to 20MB, but the OS only allocated 6.71MB

mskacelik:
https://issues.redhat.com/browse/WFLY-20678

Added a mention of the Warnings – to be consistent with the other Quickstarts.


emmartins:
@jmfinelli FYI

emmartins (approved):

PR #1048, opened 2025-08-26
Exclude mail from Kubernetes CI; Add mail OS compatibility in root RE…

emmartins:
…ADME's ToAQ


emmartins:
@Dkafetzis FYI minor fixes/enhancements on top of your work

PR #1078, opened 2025-09-09
Fix kubernetes-ci.yml that fails with empty matrix of quickstarts to …

emmartins:
…test

### DIFF
--- a/.github/workflows/kubernetes-ci.yml
+++ b/.github/workflows/kubernetes-ci.yml
@@ -154,22 +154,25 @@ jobs:
         QUICKSTARTS_TO_TEST: ${{ steps.quickstarts-to-test.outputs.quickstarts_to_test }}
       run: |
         set -x
-        matrix="{\"quickstart\":["
-        first=1
-        echo "The list of quickstarts has been determined to:"
-        for qs_dir in ${QUICKSTARTS_TO_TEST}; do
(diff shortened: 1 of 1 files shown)


emmartins:
@kabir many times the matrix of changed quickstarts gets empty, and the final step then fail, as seen in https://github.com/wildfly/quickstart/actions/runs/17589546600
To prevent this I added an if to only set the last step needed var if it's not empty (-n)
The CI won't validate this now but I have tested this with the reverse (-z) and seems to skip correctly.

.github/workflows/kubernetes-ci.yml
```diff
@@ -154,22 +154,23 @@ jobs:
...
+          done
+          matrix="${matrix}]}"
         
-        matrix="${matrix}]}"
-        
-        echo "Calculated Matrix JSON: ${matrix}"
-        echo "matrix=${matrix}" >> $GITHUB_OUTPUT
+          echo "Calculated Matrix JSON: ${matrix}"
+          echo "matrix=${matrix}" >> $GITHUB_OUTPUT
+        fi
```
kabir, line 173: Not a big deal, but perhaps a message for the else case saying "Nothing to do" or something could be useful?

emmartins:
ty @kabir

PR #1161, opened 2026-05-07
WFLY-20443 Develop a quickstart which demonstrated the dynamic SSL Cotext capability

fRandOmizer:
[WFLY-20443 Develop a quickstart which demonstrated the dynamic SSLContext capability](https://redhat.atlassian.net/browse/WFLY-20443)

### DIFF
--- a/dynamic-client-ssl-context/README-source.adoc
+++ b/dynamic-client-ssl-context/README-source.adoc
@@ -1,0 +1,99 @@
+include::../shared-doc/attributes.adoc[]
+
+= dynamic-client-ssl-context: Dynamic client SSL Context between Server and Client
+:author: Richard Záň
+:level: Intermidiate
+:technologies: Dynamic TLS, Security, Undertow
+
+[abstract]
+The dynamic-client-ssl-context quickstart shows the option to select different SSL/TLS configurations dynamically based on the host and port of the outbound connection.
+
+:standalone-server-type: default
+:archiveType: war
+
+== What is it?
+
+Elytron client supports dynamic client SSL context. It can be configured everywhere in the subsystem where client-ssl-context attribute can be configured.
+
+The `dynamic-client-ssl-context` quickstart demonstrates enabled dynamic SSL context for specified URLs and their ports. The only function they represent is to be called via provided REST client and pass as successful call through the SSL/TLS configuration.
+
+This example consists of the following two URLs, each with enabled dynamic SSL context:
+
+[cols="100%",options="headers"]
+|===
+|URL paths with their ports
+
+|`https://localhost:9443/dynamic-client-ssl-context/`
+
+|`https://localhost:10443/dynamic-client-ssl-context/`
+
+|===
+
+The server configuration is done using CLI batch scripts located in the root of the quickstart folder.
+
+// Link to the quickstart source
+include::../shared-doc/view-the-source.adoc[leveloffset=+1]
+// System Requirements
+include::../shared-doc/system-requirements.adoc[leveloffset=+1]
+// Use of {jbossHomeName}
+include::../shared-doc/use-of-jboss-home-name.adoc[leveloffset=+1]
+
+[[start_with_a_clean_server_install]]
+== Start with a Clean Server Install
+
+It is important to start with a clean version of {productName} before testing this quickstart. Make sure you unzip or install a fresh {productName} instance.
+
+// Back Up the {productName} Managed Domain Configuration
+include::../shared-doc/back-up-server-standalone-configuration.adoc[leveloffset=+1]
+// Start the {productName} Managed Domain
+include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
+
+// Build and Deploy the Quickstart
+include::../shared-doc/build-and-deploy-the-quickstart.adoc[leveloffset=+1]
+
+[[review_the_modified_server_configuration]]
+== Review the Modified Server Configuration
+
+There are too many additions to the configuration file to list here. Feel free to compare the `standalone.xml` to the backup copy to see the changes made to configure the server to run this quickstart.
+
+[[access_the_application]]
+== Access the Application
+
+The application will be running at the following URL: https://localhost:8080/dynamic-client-ssl-context
+
+A main page displays list of links:
+
+[source,options="nowrap"]
+----
+Click below to send request to different ports:
+
+Send request to port 9443
+Send request to port 10443
+----
+
+Each of links executes https GET request on server under specified ports. Each port uses a different TLS certificate. Both GET requests use the same dynamic client SSL context. For http://localhost:8080/dynamic-client-ssl-context/rest/port9443request, it will display:
+
+[source,options="nowrap"]
+----
+HTTP status of result is 200
+----
+
+Similar result is for http://localhost:8080/dynamic-client-ssl-context/rest/port10443request:
+
+[source,options="nowrap"]
+----
+HTTP status of result is 200
+----
+
+// Server Distribution Testing
+:integrationTestsDirectory: app-web/src/test/
+include::../shared-doc/run-integration-tests-with-server-distribution.adoc[leveloffset=+1]
+
+// Undeploy the Quickstart
+include::../shared-doc/undeploy-the-quickstart.adoc[leveloffset=+1]
+
+// Restore the {productName} Managed Domain Configuration Manually
+include::../shared-doc/restore-standalone-server-configuration-manual.adoc[leveloffset=+1]
+
+// Quickstart not compatible with OpenShift
+include::../shared-doc/openshift-incompatibility.adoc[leveloffset=+1]
--- a/dynamic-client-ssl-context/README.adoc
+++ b/dynamic-client-ssl-context/README.adoc
@@ -1,0 +1,424 @@
+ifdef::env-github[]
+:artifactId: dynamic-client-ssl-context
+endif::[]
+
+//***********************************************************************************
+// Enable the following flag to build README.html files for JBoss EAP product builds.
+// Comment it out for WildFly builds.
+//***********************************************************************************
+//:ProductRelease:
+
+//***********************************************************************************
+// Enable the following flag to build README.html files for EAP XP product builds.
+// Comment it out for WildFly or JBoss EAP product builds.
+//***********************************************************************************
+//:EAPXPRelease:
+
+// This is a universal name for all releases
+:ProductShortName: JBoss EAP
+// Product names and links are dependent on whether it is a product release (CD or JBoss)
+// or the WildFly project.
+// The "DocInfo*" attributes are used to build the book links to the product documentation
+
+ifdef::ProductRelease[]
+// JBoss EAP release
+:productName: JBoss EAP
+:productNameFull: Red Hat JBoss Enterprise Application Platform
+:productVersion: 8.1
+:DocInfoProductNumber: {productVersion}
+:WildFlyQuickStartRepoTag: 8.1.x
+:helmChartName: jboss-eap/eap81
+endif::[]
+
+ifdef::EAPXPRelease[]
+// JBoss EAP XP release
+:productName: JBoss EAP XP
+:productNameFull: Red Hat JBoss Enterprise Application Platform expansion pack
+:productVersion: 6.0
+:WildFlyQuickStartRepoTag: XP_6.0.0.GA
+endif::[]
+
+ifdef::ProductRelease,EAPXPRelease[]
+:githubRepoUrl: https://github.com/jboss-developer/jboss-eap-quickstarts/
+:githubRepoCodeUrl: https://github.com/jboss-developer/jboss-eap-quickstarts.git
+:jbossHomeName: EAP_HOME
+:DocInfoProductName: Red Hat JBoss Enterprise Application Platform
+:DocInfoProductNameURL: red_hat_jboss_enterprise_application_platform
+:DocInfoPreviousProductName: jboss-enterprise-application-platform
+:quickstartDownloadName: {productNameFull} {productVersion} Quickstarts
+:quickstartDownloadUrl: https://access.redhat.com/jbossnetwork/restricted/listSoftware.html?product=appplatform&downloadType=distributions
+:helmRepoName: jboss-eap
+:helmRepoUrl: https://jbossas.github.io/eap-charts/
+// END ifdef::ProductRelease,EAPXPRelease[]
+endif::[]
+
+ifndef::ProductRelease,EAPXPRelease[]
+// WildFly project
+:productName: WildFly
+:productNameFull: WildFly Application Server
+:ProductShortName: {productName}
+:jbossHomeName: WILDFLY_HOME
+:productVersion: 39
+:githubRepoUrl: https://github.com/wildfly/quickstart/
+:githubRepoCodeUrl: https://github.com/wildfly/quickstart.git
+:WildFlyQuickStartRepoTag: 39.0.0.Final
+:DocInfoProductName: Red Hat JBoss Enterprise Application Platform
+:DocInfoProductNameURL: red_hat_jboss_enterprise_application_platform
+:DocInfoPreviousProductName: jboss-enterprise-application-platform
+:helmRepoName: wildfly
+:helmRepoUrl: http://docs.wildfly.org/wildfly-charts/
+:helmChartName: wildfly/wildfly
+// END ifndef::ProductRelease,EAPCDRelease,EAPXPRelease[]
+endif::[]
+
+:source: {githubRepoUrl}
+
+// Values for Openshift S2i sections attributes
+:EapForOpenshiftBookName: {productNameFull} for OpenShift
+:EapForOpenshiftOnlineBookName: {EapForOpenshiftBookName} Online
+:xpaasproduct: {productNameFull} for OpenShift
+:xpaasproduct-shortname: {ProductShortName} for OpenShift
+:ContainerRegistryName: Red Hat Container Registry
+:EapForOpenshiftBookName: Getting Started with {ProductShortName} for OpenShift Container Platform
+:EapForOpenshiftOnlineBookName: Getting Started with {ProductShortName} for OpenShift Online
+:OpenShiftOnlinePlatformName: Red Hat OpenShift Container Platform
+:OpenShiftOnlineName: Red Hat OpenShift Online
+// Links to the OpenShift documentation
+:LinkOpenShiftGuide: https://access.redhat.com/documentation/en-us/{DocInfoProductNameURL}/{DocInfoProductNumber}/html-single/getting_started_with_jboss_eap_for_openshift_container_platform/
+:LinkOpenShiftOnlineGuide: https://access.redhat.com/documentation/en-us/{DocInfoProductNameURL}/{DocInfoProductNumber}/html-single/getting_started_with_jboss_eap_for_openshift_online/
+
+ifdef::EAPXPRelease[]
+// Attributes for XP releases
+:EapForOpenshiftBookName: {productNameFull} for OpenShift
+:EapForOpenshiftOnlineBookName: {productNameFull} for OpenShift Online
+:xpaasproduct: {productNameFull} for OpenShift
+:ContainerRegistryName: Red Hat Container Registry
+:EapForOpenshiftBookName: {productNameFull} for OpenShift
+:EapForOpenshiftOnlineBookName: {productNameFull} for OpenShift Online
+// Links to the OpenShift documentation
+:LinkOpenShiftGuide: https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/{DocInfoProductNumber}/html/using_eclipse_microprofile_in_jboss_eap/using-the-openshift-image-for-jboss-eap-xp_default
+:LinkOpenShiftOnlineGuide: https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/{DocInfoProductNumber}/html/using_eclipse_microprofile_in_jboss_eap/using-the-openshift-image-for-jboss-eap-xp_default
+endif::[]
+
+//*************************
+// Other values
+//*************************
+:buildRequirements: Java SE 17.0 or later, and Maven 3.6.0 or later
+:javaVersion: Jakarta EE 10
+ifdef::EAPXPRelease[]
+:javaVersion: Eclipse MicroProfile
+endif::[]
+:guidesBaseUrl: https://github.com/jboss-developer/jboss-developer-shared-resources/blob/master/guides/
+:useEclipseUrl: {guidesBaseUrl}USE_JBDS.adoc#use_red_hat_jboss_developer_studio_or_eclipse_to_run_the_quickstarts
+:useEclipseDeployJavaClientDocUrl: {guidesBaseUrl}USE_JBDS.adoc#deploy_and_undeploy_a_quickstart_containing_server_and_java_client_projects
+:useEclipseDeployEARDocUrl: {guidesBaseUrl}USE_JBDS.adoc#deploy_and_undeploy_a_quickstart_ear_project
+:useProductHomeDocUrl: {guidesBaseUrl}USE_OF_{jbossHomeName}.adoc#use_of_product_home_and_jboss_home_variables
+:configureMavenDocUrl: {guidesBaseUrl}CONFIGURE_MAVEN_JBOSS_EAP.adoc#configure_maven_to_build_and_deploy_the_quickstarts
+:addUserDocUrl: {guidesBaseUrl}CREATE_USERS.adoc#create_users_required_by_the_quickstarts
+:addApplicationUserDocUrl: {guidesBaseUrl}CREATE_USERS.adoc#add_an_application_user
+:addManagementUserDocUrl: {guidesBaseUrl}CREATE_USERS.adoc#add_an_management_user
+:startServerDocUrl: {guidesBaseUrl}START_JBOSS_EAP.adoc#start_the_jboss_eap_server
+:configurePostgresDocUrl: {guidesBaseUrl}CONFIGURE_POSTGRESQL_JBOSS_EAP.adoc#configure_the_postgresql_database_for_use_with_the_quickstarts
+:configurePostgresDownloadDocUrl: {guidesBaseUrl}CONFIGURE_POSTGRESQL_JBOSS_EAP.adoc#download_and_install_postgresql
+:configurePostgresCreateUserDocUrl: {guidesBaseUrl}CONFIGURE_POSTGRESQL_JBOSS_EAP.adoc#create_a_database_user
+:configurePostgresAddModuleDocUrl: {guidesBaseUrl}CONFIGURE_POSTGRESQL_JBOSS_EAP.adoc#add_the_postgres_module_to_the_jboss_eap_server
+:configurePostgresDriverDocUrl: {guidesBaseUrl}CONFIGURE_POSTGRESQL_JBOSS_EAP.adoc#configure_the_postgresql_driver_in_the_jboss_eap_server
+:configureBytemanDownloadDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#download_and_configure_byteman
+:configureBytemanDisableDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#disable_the_byteman_script
+:configureBytemanClearDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#clear_the_transaction_object_store
+:configureBytemanQuickstartDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#configure_byteman_for_use_with_the_quickstarts
+:configureBytemanHaltDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#use_byteman_to_halt_the_application[
+:configureBytemanQuickstartsDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#configure_byteman_for_use_with_the_quickstarts
+
+= dynamic-client-ssl-context: Dynamic client SSL Context between Server and Client
+:author: Richard Záň
+:level: Intermidiate
+:technologies: Dynamic TLS, Security, Undertow
+
+[abstract]
+The dynamic-client-ssl-context quickstart shows the option to select different SSL/TLS configurations dynamically based on the host and port of the outbound connection.
+
+:standalone-server-type: default
+:archiveType: war
+
+== What is it?
+
+Elytron client supports dynamic client SSL context. It can be configured everywhere in the subsystem where client-ssl-context attribute can be configured.
+
+The `dynamic-client-ssl-context` quickstart demonstrates enabled dynamic SSL context for specified URLs and their ports. The only function they represent is to be called via provided REST client and pass as successful call through the SSL/TLS configuration.
+
+This example consists of the following two URLs, each with enabled dynamic SSL context:
+
+[cols="100%",options="headers"]
+|===
+|URL paths with their ports
+
+|`https://localhost:9443/dynamic-client-ssl-context/`
+
+|`https://localhost:10443/dynamic-client-ssl-context/`
+
+|===
+
+The server configuration is done using CLI batch scripts located in the root of the quickstart folder.
+
+// Link to the quickstart source
+:leveloffset: +1
+
+ifndef::ProductRelease,EAPXPRelease[]
+link:https://github.com/wildfly/quickstart/tree/{WildFlyQuickStartRepoTag}/{artifactId}[Browse the source]
+endif::[]
+
+:leveloffset!:
+// System Requirements
+:leveloffset: +1
+
+[[system_requirements]]
+= System Requirements
+
+The application this project produces is designed to be run on {productNameFull} {productVersion} or later.
+
+All you need to build this project is {buildRequirements}. See link:{configureMavenDocUrl}[Configure Maven to Build and Deploy the Quickstarts] to make sure you are configured correctly for testing the quickstarts.
+
+:leveloffset!:
+// Use of {jbossHomeName}
+:leveloffset: +1
+
+ifdef::requires-multiple-servers[]
+[[use_of_jboss_home_name]]
+= Use of the {jbossHomeName}_1, {jbossHomeName}_2, and QUICKSTART_HOME Variables
+
+This quickstart requires that you clone your `__{jbossHomeName}__` installation directory and run two servers. The installation path is described in detail here: link:{useProductHomeDocUrl}[Use of __{jbossHomeName}__ and __JBOSS_HOME__ Variables].
+
+In the following instructions, replace `__{jbossHomeName}_1__` with the path to your first {productName} server and replace `__{jbossHomeName}_2__` with the path to your second cloned {productName} server.
+
+When you see the replaceable variable __QUICKSTART_HOME__, replace it with the path to the root directory of all of the quickstarts.
+endif::[]
+
+ifdef::optional-domain-or-multiple-servers[]
+[[use_of_jboss_home_name]]
+= Use of the {jbossHomeName}_1, {jbossHomeName}_2, and QUICKSTART_HOME Variables
+
+When deploying this quickstart to a managed domain, replace `__{jbossHomeName}__` with the actual path to your {productName} installation. The installation path is described in detail here: link:{useProductHomeDocUrl}[Use of __{jbossHomeName}__ and __JBOSS_HOME__ Variables].
+
+When deploying this quickstart to multiple standalone servers, this quickstart requires that you clone your `__{jbossHomeName}__` installation directory and run two servers. In the following instructions, replace `__{jbossHomeName}_1__` with the path to your first {productName} server and replace `__{jbossHomeName}_2__` with the path to your second cloned {productName} server.
+
+When you see the replaceable variable __QUICKSTART_HOME__, replace it with the path to the root directory of all of the quickstarts.
+endif::[]
+
+ifndef::requires-multiple-servers,optional-domain-or-multiple-servers[]
+[[use_of_jboss_home_name]]
+= Use of the {jbossHomeName} and QUICKSTART_HOME Variables
+
+In the following instructions, replace `__{jbossHomeName}__` with the actual path to your {productName} installation. The installation path is described in detail here: link:{useProductHomeDocUrl}[Use of __{jbossHomeName}__ and __JBOSS_HOME__ Variables].
+
+When you see the replaceable variable __QUICKSTART_HOME__, replace it with the path to the root directory of all of the quickstarts.
+endif::[]
+
+:leveloffset!:
+
+[[start_with_a_clean_server_install]]
+== Start with a Clean Server Install
+
+It is important to start with a clean version of {productName} before testing this quickstart. Make sure you unzip or install a fresh {productName} instance.
+
+// Back Up the {productName} Managed Domain Configuration
+:leveloffset: +1
+
+[[back_up_standalone_server_configuration]]
+= Back Up the {productName} Standalone Server Configuration
+
+Before you begin, back up your server configuration file.
+
+. If it is running, stop the {productName} server.
+. Back up the `{jbossHomeName}/standalone/configuration/standalone.xml` file.
+
+After you have completed testing this quickstart, you can replace this file to restore the server to its original configuration.
+
+:leveloffset!:
+// Start the {productName} Managed Domain
+:leveloffset: +1
+
+[[start_the_eap_standalone_server]]
+= Start the {productName} Standalone Server
+
+. Open a terminal and navigate to the root of the {productName} directory.
+. Start the {productName} server with the {serverProfile} by typing the following command.
++
+[source,subs="+quotes,attributes+",options="nowrap"]
+----
+$ __{jbossHomeName}__/bin/standalone.sh
+----
++
+NOTE: For Windows, use the `__{jbossHomeName}__\bin\standalone.bat` script.
+
+
+:leveloffset!:
+
+// Build and Deploy the Quickstart
+:leveloffset: +1
+
+[[build_and_deploy_the_quickstart]]
+= Build and Deploy the Quickstart
+
+// The archive name defaults to the artifactId if not overridden
+ifndef::archiveName[]
+:archiveName: {artifactId}
+endif::archiveName[]
+
+// The archive type defaults to war if not overridden
+ifndef::archiveType[]
+:archiveType: war
+endif::archiveType[]
+
+// Define the archive file name as the concatenation of "archiveName" + "." + "archiveType+
+:archiveFileName: {archiveName}.{archiveType}
+
+// If they have not defined the target archive directory, make it the default for the archive type.
+ifndef::archiveDir[]
+
+ifeval::["{archiveType}"=="ear"]
+:archiveDir: {artifactId}/ear/target
+endif::[]
+
+ifeval::["{archiveType}"=="war"]
+:archiveDir: {artifactId}/target
+endif::[]
+
+ifeval::["{archiveType}"=="jar"]
+:archiveDir: {artifactId}/target
+endif::[]
+
+endif::archiveDir[]
+
+ifndef::mavenCommand[]
+ifeval::["{archiveType}"=="ear"]
+:mavenCommand: clean install
+endif::[]
+
+ifeval::["{archiveType}"=="war"]
+:mavenCommand: clean package
+endif::[]
+
+ifeval::["{archiveType}"=="jar"]
+:mavenCommand: clean install
+endif::[]
+
+endif::mavenCommand[]
+
+. Make sure {productName} server is started.
+. Open a terminal and navigate to the root directory of this quickstart.
+. Type the following command to build the quickstart.
++
+[source,subs="attributes+",options="nowrap"]
+----
+$ mvn {mavenCommand}
+----
+
+. Type the following command to deploy the quickstart.
++
+[source,subs="attributes+",options="nowrap"]
+----
+$ mvn wildfly:deploy
+----
+
+This deploys the `{archiveDir}/{archiveFileName}` to the running instance of the server.
+
+You should see a message in the server log indicating that the archive deployed successfully.
+
+:leveloffset!:
+
+[[review_the_modified_server_configuration]]
+== Review the Modified Server Configuration
+
+There are too many additions to the configuration file to list here. Feel free to compare the `standalone.xml` to the backup copy to see the changes made to configure the server to run this quickstart.
+
+[[access_the_application]]
+== Access the Application
+
+The application will be running at the following URL: https://localhost:8080/dynamic-client-ssl-context
+
+A main page displays list of links:
+
+[source,options="nowrap"]
+----
+Click below to send request to different ports:
+
+Send request to port 9443
+Send request to port 10443
+----
+
+Each of links executes https GET request on server under specified ports. Each port uses a different TLS certificate. Both GET requests use the same dynamic client SSL context. For http://localhost:8080/dynamic-client-ssl-context/rest/port9443request, it will display:
+
+[source,options="nowrap"]
+----
+HTTP status of result is 200
+----
+
+Similar result is for http://localhost:8080/dynamic-client-ssl-context/rest/port10443request:
+
+[source,options="nowrap"]
+----
+HTTP status of result is 200
+----
+
+// Server Distribution Testing
+:integrationTestsDirectory: app-web/src/test/
+:leveloffset: +1
+
+[[run_the_integration_tests_with_server_distribution]]
+= Run the Integration Tests
+ifndef::integrationTestsDirectory[:integrationTestsDirectory: src/test/]
+ifndef::extraStandardDistTestParams[:extraStandardDistTestParams: ]
+
+This quickstart includes integration tests, which are located under the `{integrationTestsDirectory}` directory. The integration tests verify that the quickstart runs correctly when deployed on the server.
+
+Follow these steps to run the integration tests.
+
+. Make sure {productName} server is started.
+. Make sure the quickstart is deployed.
+. Type the following command to run the `verify` goal with the `integration-testing` profile activated.
++
+[source,subs="attributes+",options="nowrap"]
+----
+$ mvn verify -Pintegration-testing {extraStandardDistTestParams}
+----
+
+:leveloffset!:
+
+// Undeploy the Quickstart
+:leveloffset: +1
+
+[[undeploy_the_quickstart]]
+= Undeploy the Quickstart
+
+When you are finished testing the quickstart, follow these steps to undeploy the archive.
+
+. Make sure {productName} server is started.
+. Open a terminal and navigate to the root directory of this quickstart.
+. Type this command to undeploy the archive:
++
+[source,options="nowrap"]
+----
+$ mvn wildfly:undeploy
+----
+
+:leveloffset!:
+
+// Restore the {productName} Managed Domain Configuration Manually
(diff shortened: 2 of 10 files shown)


fRandOmizer:
@skyllarr , please. take a look.

helloworld-dynamic-ssl-context/configure-dynamic-ssl-context.cli
```diff
@@ -0,0 +1,90 @@
+# Script to configure dynamic SSL context in the JBoss EAP server
+
+# Create the server's keystores
+/subsystem=elytron/key-store=twoWayKS1:add(path=server1.keystore.P12,relative-to=jboss.server.config.dir,credential-reference={clear-text=secret},type=JKS)
```
skyllarr: Let's not use JKS anywhere, it's becoming obsolete, we should use PKCS12

helloworld-dynamic-ssl-context/README.adoc
```diff
@@ -0,0 +1,490 @@
...
+ifeval::["{archiveType}"=="jar"]
+:mavenCommand: clean install
+endif::[]
+
+endif::mavenCommand[]
+
+. Make sure {productName} server is started.
+. Open a terminal and navigate to the root directory of this quickstart.
+ifdef::reactive-messaging[]
+. Run this command to enable the MicroProfile Reactive Messaging functionality on the server
```
skyllarr: Let's not forget to remove this readme

dynamic-client-ssl-context/src/main/webapp/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
```
skyllarr, line 5: missing header

helloworld-dynamic-ssl-context/src/test/java/org/jboss/as/quickstarts/helloworld_dynamic_ssl_context/BasicRuntimeIT.java
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright 2023 JBoss by Red Hat.
```
skyllarr: this is old version of header

dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/BasicRuntimeIT.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * The very basic runtime integration testing.
+ * @author Richard Zan
+ */
+public class BasicRuntimeIT {
+
+    private static final String SERVER_HOST_FIRST_PORT = "https://localhost:9443";
+    private static final String SERVER_HOST_SECOND_PORT = "https://localhost:10443";
+    private static final String ARTIFACT_ID = "/helloworld-dynamic-ssl-context/";
+    private static final String DEFAULT_SERVER_DIR = System.getProperty("user.dir") + "/target/server";
```
skyllarr, line 43: Can it be JBOSS_HOME? We often use it

helloworld-dynamic-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+[abstract]
+The `helloworld-dynamic-ssl-context` quickstart shows the option to select different SSL/TLS configurations dynamically based on the host and port of the outbound connection by the Elytron subsystem.
+
+:standalone-server-type: default
+:archiveType: war
+
+== What is it?
+
+The `helloworld-dynamic-ssl-context` quickstart demonstrates enabled dynamic ssl context for specified URLs and their ports. The only function they represents is to be called via provided REST client and pass as successful call through the SSL/TLS configuration.
```
skyllarr: I would mention somewhere here that elytron client also supports dynamic ssl context and that it can be configured anywhere in the subsystem where ssl-context attribute can be configured. SO it is not just with the REST client

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+The server configuration is done using CLI batch scripts located in the root of the quickstart folder.
+
+// Link to the quickstart source
+include::../shared-doc/view-the-source.adoc[leveloffset=+1]
+// System Requirements
+include::../shared-doc/system-requirements.adoc[leveloffset=+1]
+// Use of {jbossHomeName}
+include::../shared-doc/use-of-jboss-home-name.adoc[leveloffset=+1]
+
```
skyllarr, line 40: we can mention how to run them here to be consistent with the rest of quickstarts

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,98 @@
...
+// Start the {productName} Managed Domain
+include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
+
+// Build and Deploy the Quickstart
+include::../shared-doc/build-and-deploy-the-quickstart.adoc[leveloffset=+1]
+
+[[review_the_modified_server_configuration]]
+== Review the Modified Server Configuration
+
+There are too many additions to the configuration file to list here. Feel free to compare the `standalone.xml` to the backup copy to see the changes made to configure the server to run this quickstart.
```
skyllarr, line 57: I am lazy to check, but is this sentence consistent with some other quickstarts?

helloworld-dynamic-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+[source,options="nowrap"]
+----
+Click below to send request to different ports:
+
+Send request to port 9443
+Send request to port 10443
+----
+
+Each of links executes https GET request on server under specified ports, which has enabled dynamic SSL context.
```
skyllarr: I would mention that it is the same ssl context configured

skyllarr (commented):
Hi @fRandOmizer , I just added some initial minor comments,  I will try to run this quickstart once these are fixed to make sure it works.

dynamic-client-ssl-context/README-source.adoc
fRandOmizer: yes, I have took the liberty to reuse this one:

> There are too many additions to the configuration files to list here. Feel free to compare the `domain.xml` and `host.xml` to the backup copies to see the changes made to configure the server to run this quickstart.

https://github.com/wildfly/quickstart/blob/main/ejb-multi-server/README.adoc?plain=1#L349

dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/BasicRuntimeIT.java
fRandOmizer: I think, it is not a good idea, this test should run on clean WildFly instance with the this Quickstart.
Therefore, JBOSS_HOME should not be pointed at as customer WildFly should not contain the quickstart, which would cause the test to fail.
The whole reason as why I want them to have a clean WildFly, is the fact that, we are adding endpoints to the configuration and heavily modifing them, which could conflict with the already running customer WildFly instance that possibly uses these ports.

dynamic-client-ssl-context/README-source.adoc
fRandOmizer: something like this? https://github.com/wildfly/quickstart/blob/main/helloworld-mutual-ssl-secured/README.adoc#building-and-running-the-quickstart-application-with-provisioned-wildfly-server

I have already added parts that cover it:
`../shared-doc/start-the-standalone-server.adoc`
`../shared-doc/build-and-deploy-the-quickstart.adoc`
`../shared-doc/run-integration-tests-with-server-distribution.adoc`
...

I took a massive inspiration from https://github.com/wildfly/quickstart/blob/main/ejb-multi-server/README-source.adoc

dynamic-client-ssl-context/src/main/webapp/index.html
fRandOmizer: what you exactly mean? `<head></head>` element is defined, I have added `<h1></h1>` title

helloworld-dynamic-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,100 @@
+include::../shared-doc/attributes.adoc[]
+
+= helloworld-dynamic-ssl-context: Dynamic SSL Context between Server and Client
```
skyllarr: just a nitpick, it is only client ssl context, we can name this:

helloworld-dynamic-client-ssl-context: Dynamic client SSL Context

helloworld-dynamic-ssl-context/README.adoc
```diff
@@ -0,0 +1,484 @@
...
+:configurePostgresAddModuleDocUrl: {guidesBaseUrl}CONFIGURE_POSTGRESQL_JBOSS_EAP.adoc#add_the_postgres_module_to_the_jboss_eap_server
+:configurePostgresDriverDocUrl: {guidesBaseUrl}CONFIGURE_POSTGRESQL_JBOSS_EAP.adoc#configure_the_postgresql_driver_in_the_jboss_eap_server
+:configureBytemanDownloadDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#download_and_configure_byteman
+:configureBytemanDisableDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#disable_the_byteman_script
+:configureBytemanClearDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#clear_the_transaction_object_store
+:configureBytemanQuickstartDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#configure_byteman_for_use_with_the_quickstarts
+:configureBytemanHaltDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#use_byteman_to_halt_the_application[
+:configureBytemanQuickstartsDocUrl: {guidesBaseUrl}CONFIGURE_BYTEMAN.adoc#configure_byteman_for_use_with_the_quickstarts
+
+= helloworld-dynamic-ssl-context: Dynamic SSL Context between Server and Client
```
skyllarr: helloworld-dynamic-ssl-context: Dynamic client SSL Context

helloworld-dynamic-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,100 @@
...
+:author: Richard Záň
+:level: Intermidiate
+:technologies: Dynamic TLS, Security, Undertow
+
+[abstract]
+The `helloworld-dynamic-ssl-context` quickstart shows the option to select different SSL/TLS configurations dynamically based on the host and port of the outbound connection by the Elytron subsystem.
```
skyllarr: The `helloworld-dynamic-ssl-context` quickstart shows the option to select different SSL/TLS configurations dynamically based on the host and port of the outbound connection.

I removed the ending to make it easier to read

helloworld-dynamic-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+:standalone-server-type: default
+:archiveType: war
+
+== What is it?
+
+Elytron client supports dynamic ssl context. It can be configured everywhere in the subsystems similarly as ssl-context attribute.
```
skyllarr: Elytron client supports dynamic client SSL context. It can be configured everywhere in the subsystems similarly as client-ssl-context attribute.

helloworld-dynamic-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+The `helloworld-dynamic-ssl-context` quickstart demonstrates enabled dynamic ssl context for specified URLs and their ports. The only function they represent is to be called via provided REST client and pass as successful call through the SSL/TLS configuration.
+
+This example consists of the following two urls, each with enabled dynamic ssl context:
```
skyllarr: Let's unify how we write URL / url and SSL context / ssl context. So it is the same in all text

helloworld-dynamic-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+[source,options="nowrap"]
+----
+Click below to send request to different ports:
+
+Send request to port 9443
+Send request to port 10443
+----
+
+Each of links executes https GET request on server under specified ports, which has enabled the same dynamic SSL context.
```
skyllarr: Each of links executes https GET request on server under specified ports. Each port uses a different TLS certificate. Both GET requests use the same dynamic client SSL context. For http://localhost:8080/helloworld-dynamic-ssl-context/rest/port9443request, it will display:

helloworld-dynamic-ssl-context/README.adoc
```diff
@@ -0,0 +1,484 @@
...
+link:https://github.com/wildfly/quickstart/tree/{WildFlyQuickStartRepoTag}/{artifactId}[Browse the source]
+endif::[]
+
+:leveloffset!:
+// System Requirements
+:leveloffset: +1
+
+[[system_requirements]]
+= System Requirements
+//******************************************************************************
```
skyllarr: is this file needed? since it is commented out

helloworld-dynamic-ssl-context/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+        <maven.compiler.release>17</maven.compiler.release>
+        <!-- the version for the Server -->
+        <version.server>39.0.0.Final</version.server>
+        <!-- the versions for BOMs, Packs and Plugins -->
+        <version.bom.ee>${version.server}</version.bom.ee>
+        <version.plugin.wildfly>5.1.5.Final</version.plugin.wildfly>
+    </properties>
+
+    <artifactId>helloworld-dynamic-ssl-context</artifactId>
+    <version>40.0.0.Beta1-SNAPSHOT</version>
```
skyllarr: above is version 39 and here is 40, is that correct?

helloworld-dynamic-ssl-context/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+    <packaging>war</packaging>
+    <name>Quickstart: helloworld-dynamic-ssl-context</name>
+    <description>Helloworld Dynamic SSL Context</description>
```
skyllarr: I would rename this quickstat helloworld-dynamic-client-ssl-context and describe Hello world dynamic client SSL context

dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
```diff
@@ -0,0 +1,90 @@
...
+# Add https listeners
+/subsystem=undertow/server=default-server/https-listener=first-listener:add(socket-binding=first-socket-binding,ssl-context=twoWaySSC1,enable-http2=true)
+/subsystem=undertow/server=default-server/https-listener=second-listener:add(socket-binding=second-socket-binding,ssl-context=twoWaySSC2,enable-http2=true)
+
+# Add the authentication authentication context
+/subsystem=elytron/authentication-context=a:add(match-rules=[{match-port=9443,ssl-context=client1-ssl-context},{match-port=10443,ssl-context=client2-ssl-context}])
+
+# Set up dynamic client SSL context
+/subsystem=elytron/dynamic-client-ssl-context=dynamicClientSSLContext:add(authentication-context=a)
+/subsystem=elytron:write-attribute(name=default-ssl-context,value=dynamicClientSSLContext)
```
skyllarr, line 92: Let's add a comment that we configure the dynamic client SSL context to be a default SSL Context for the whole server.

skyllarr (commented):
I just added some minors. Thanks @fRandOmizer !

helloworld-dynamic-ssl-context/README-source.adoc
fRandOmizer: Okay, I will update just the second part of the readme. Otherwise I would have to rename the whole project.

helloworld-dynamic-ssl-context/README-source.adoc
fRandOmizer: It will be URL/URLs and SSL.

helloworld-dynamic-ssl-context/README.adoc
fRandOmizer: Honestly, I don't know. It is my first encounter with .adoc format and the way QuickStart it is using. I can delete it without problem.

fRandOmizer: I have deleted almost all commented blocks of text.

helloworld-dynamic-ssl-context/pom.xml
fRandOmizer: I had to update the project. It was correct.
Here is an example: https://github.com/wildfly/quickstart/blob/main/helloworld-mutual-ssl-secured/pom.xml#L33 and https://github.com/wildfly/quickstart/blob/main/helloworld-mutual-ssl-secured/pom.xml#L50

helloworld-dynamic-ssl-context/pom.xml
fRandOmizer: okay... I will rename the whole project

helloworld-dynamic-client-ssl-context/src/main/java/org/jboss/as/quickstarts/helloworld/HelloService.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2015, Red Hat, Inc. and/or its affiliates, and individual
```
rhusar: My watch says 2026? :-D

rhusar (changes requested):

helloworld-dynamic-client-ssl-context/src/main/java/org/jboss/as/quickstarts/helloworld/HelloService.java
fRandOmizer: the problem is, there is like dozens of CopyRights(CR) and each have earlier date which does not match with initial commit date... if you have the most up to date CR, I will use it in all classes

darranl: If it is a new file use the current year, if you are editing an existing file don't touch the date.

fRandOmizer: ok, I will update the year, the rest of the CR is up to date?

fRandOmizer: updated

rhusar: @emmartins Isn't this project under commonhouse? I am pretty sure it is, so the copyright notice is wrong, no?

helloworld-dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
+include::../shared-doc/attributes.adoc[]
+
+= helloworld-dynamic-client-ssl-context: Dynamic client SSL Context between Server and Client
```
skyllarr: Not sure why some QSs start with helloworld, but I would rename to just `dynamic-client-ssl-context` for simplicity and it's a bit shorter. The `helloworld` word seems redundant

dynamic-client-ssl-context/src/main/webapp/index.html
skyllarr: I meant copyright header, like here for example https://github.com/wildfly/quickstart/blob/main/ejb-remote/src/main/webapp/index.html

dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/BasicRuntimeIT.java
skyllarr: Sorry not JBOSS_HOME but WILDFLY_HOME. In the README for this quickstart here https://github.com/fRandOmizer/quickstart/tree/WFLY-20443/helloworld-dynamic-client-ssl-context it mentions to start with a clean server install. And it mentions WILDFLY_HOME variable.

rhusar: @skyllarr Please don't use WILDFLY_HOME - https://redhat.atlassian.net/browse/WFLY-22020

helloworld-dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/helloworld_dynamic_client_ssl_context/KeyStoreUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.FileNotFoundException;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.cert.Certificate;
+
+/**
+ * An interface to obtain the certificate from a keystore file and create s truststore.
```
skyllarr: typo `s truststore`

helloworld-dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
```diff
@@ -0,0 +1,90 @@
+# Script to configure dynamic SSL context in the JBoss EAP server
```
skyllarr: s/JBoss EAP/WildFly

helloworld-dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
```diff
@@ -0,0 +1,90 @@
...
+# Configure the "first" and "second" connections for the main server
+# Add the socket binding to the full-sockets, used by the app-main server
+/socket-binding-group=standard-sockets/socket-binding=first-socket-binding:add(port=9443)
+/socket-binding-group=standard-sockets/socket-binding=second-socket-binding:add(port=10443)
+
+# Add https listeners
+/subsystem=undertow/server=default-server/https-listener=first-listener:add(socket-binding=first-socket-binding,ssl-context=twoWaySSC1,enable-http2=true)
+/subsystem=undertow/server=default-server/https-listener=second-listener:add(socket-binding=second-socket-binding,ssl-context=twoWaySSC2,enable-http2=true)
+
+# Add the authentication authentication context
```
skyllarr: s/authentication authentication context/authentication context

helloworld-dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
```diff
@@ -0,0 +1,90 @@
...
+/subsystem=elytron/authentication-context=a:add(match-rules=[{match-port=9443,ssl-context=client1-ssl-context},{match-port=10443,ssl-context=client2-ssl-context}])
+
+# Set up dynamic client SSL context to be a default SSL Context for the whole server
+/subsystem=elytron/dynamic-client-ssl-context=dynamicClientSSLContext:add(authentication-context=a)
```
skyllarr: Maybe we can rename `a` for `ac` at least for the QS

helloworld-dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+[abstract]
+The helloworld-dynamic-client-ssl-context quickstart shows the option to select different SSL/TLS configurations dynamically based on the host and port of the outbound connection.
+
+:standalone-server-type: default
+:archiveType: war
+
+== What is it?
+
+Elytron client supports dynamic client SSL context. It can be configured everywhere in the subsystems similarly as client-ssl-context attribute.
```
skyllarr: s/It can be configured everywhere in the subsystems similarly as client-ssl-context attribute./It can be configured everywhere in the subsystem where client-ssl-context attribute can be configured.

skyllarr:
@fRandOmizer I've added some nitpicks otherwise looks good, thank you!

helloworld-dynamic-client-ssl-context/src/main/java/org/jboss/as/quickstarts/helloworld/HelloService.java
emmartins: @rhusar yes the copyright should be same as in WildFly:

Java:
```
/*
 */
```

XML:
```
<!--
  ~ Copyright The WildFly Authors
  ~ SPDX-License-Identifier: Apache-2.0
  -->
```

Text:
```
#
# Copyright The WildFly Authors
# SPDX-License-Identifier: Apache-2.0
#
```

helloworld-dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
emmartins: Please avoid using WildFly or JBoss EAP, instead just use "server" or "application server". In README-source.adoc please use {productName}.

dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/BasicRuntimeIT.java
emmartins: All QuickStarts should try to reuse the common integration tests setup, otherwise you can't reuse shared-docs content. Taking this into account please try to use the server.host/SERVER_HOST properties only, with full address to the app (e.g. http://localhost:8080/helloworld) and for that infer any other params you need, for instance you can parse the port and then add 1000 to calculate the second port...

fRandOmizer: hopefully resolved

dynamic-client-ssl-context/src/main/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/Endpoint.java
```diff
@@ -0,0 +1,48 @@
...
+    private ResteasyClient client = builder.hostnameVerifier((s, sslSession) -> true).sslContext(SSLContext.getDefault()).build(); // resteasy client must set default ssl context otherwise it uses null
+
+    public Endpoint() throws NoSuchAlgorithmException {
+    }
+
+    @GET
+    @Path("/port9443request")
+    @Produces(MediaType.TEXT_HTML)
+    public String pingFirstServer() {
+        Response response = client.target("https://127.0.0.1:9443/dynamic-client-ssl-context/").request().get();
```
emmartins: this host should not be hardcoded

dynamic-client-ssl-context/src/main/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/Endpoint.java
```diff
@@ -0,0 +1,48 @@
...
+        return "HTTP status of result is " +  response.getStatus();
+    }
+
+    @GET
+    @Path("/port10443request")
+    @Produces(MediaType.TEXT_HTML)
+    public String pingSecondServer() throws IOException {
+        Response response = client.target("https://127.0.0.1:10443/dynamic-client-ssl-context/").request().get();
```
emmartins: this host should not be hardcoded

dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/BasicRuntimeIT.java
emmartins: @fRandOmizer I was thinking something like this:

```
private static final String DEFAULT_SERVER_HOST = "https://localhost:9443";
private static final int SECOND_PORT_OFFSET = 1000;

private String getFirstServerHost() {
        String serverHost = System.getenv("SERVER_HOST");
        if (serverHost == null) {
            serverHost = System.getProperty("server.host");
        }
        if (serverHost == null) {
            serverHost = DEFAULT_SERVER_HOST;
        }
        serverHost += ARTIFACT_ID;
        return serverHost;
    }

private String getSecondServerHost() {
        String serverHost = getFirstServerHost();
        URI uri = URI.create(url);
        int port = uri.getPort();
        if (port == -1) {
                port = "https".equalsIgnoreCase(uri.getScheme()) ? 443 : 80;
         }
        return String.format("%s://%s:%d%s",
            uri.getScheme(),
            uri.getHost(),
            port + SECOND_PORT_OFFSET,
            uri.getPath());
}

```

fRandOmizer: @emmartins The ports are hardcoded in the cli script on line 78-81:

// Configure the "first" and "second" connections for the main server
// Add the socket binding to the full-sockets, used by the app-main server

/socket-binding-group=standard-sockets/socket-binding=first-socket-binding:add(port=9443)
/socket-binding-group=standard-sockets/socket-binding=second-socket-binding:add(port=10443)


Your approach however ignores the fact that the default port is 8443 as it is in test https://github.com/wildfly/quickstart/blob/main/helloworld-mutual-ssl-secured/src/test/java/org/jboss/as/quickstarts/helloworld_mutual_ssl_secured/BasicRuntimeIT.java#L55

Therefore, I have to edit both urls with predefined ports.

However, I will add validation of url. Even if it should not be done by this test.

fRandOmizer: done

dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
```diff
@@ -0,0 +1,95 @@
...
+
+# Add the authentication context
+/subsystem=elytron/authentication-context=ac:add(match-rules=[{match-port=9443,ssl-context=client1-ssl-context},{match-port=10443,ssl-context=client2-ssl-context}])
+
+# Set up dynamic client SSL context to be a default SSL Context for the whole server
+/subsystem=elytron/dynamic-client-ssl-context=dynamicClientSSLContext:add(authentication-context=ac)
+/subsystem=elytron:write-attribute(name=default-ssl-context,value=dynamicClientSSLContext)
+
+# Reload the server configuration
+reload
```
emmartins, line 95: please remove the reload, it fails when executed by wildfly maven plugin

dynamic-client-ssl-context/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+            but on testing clients like this we could use commons-logging instead -->
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>commons-logging-jboss-logging</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-client-api</artifactId>
+            <version>7.0.2.Final</version>
```
emmartins, line 106: depend instead on resteasy-client, remove the version (it is managed by the imported BOM), and change scope to provided (otherwise it gets bundled in the deployment)

dynamic-client-ssl-context/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.wildfly.plugins</groupId>
+                        <artifactId>wildfly-maven-plugin</artifactId>
+                        <configuration>
+                            <feature-packs>
```
emmartins, line 137: why not use Glow auto discovery of feature packs and layers, like all other quickstarts?

dynamic-client-ssl-context/README.adoc
```diff
@@ -0,0 +1,424 @@
+ifdef::env-github[]
```
emmartins, line 1: please do not include a README.adoc, that is the flattened version of README-source.adoc, which is GitHub friendly and is auto built by one of our GitHub workflows

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+This example consists of the following two URLs, each with enabled dynamic SSL context:
+
+[cols="100%",options="headers"]
+|===
+|URL paths with their ports
+
+|`https://localhost:9443/dynamic-client-ssl-context/`
+
+|`https://localhost:10443/dynamic-client-ssl-context/`
```
emmartins, line 28: IMHO this doesn't belong to an abstract / "what is it" section, and should be removed

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+The server configuration is done using CLI batch scripts located in the root of the quickstart folder.
+
+// Link to the quickstart source
+include::../shared-doc/view-the-source.adoc[leveloffset=+1]
+// System Requirements
+include::../shared-doc/system-requirements.adoc[leveloffset=+1]
+// Use of {jbossHomeName}
+include::../shared-doc/use-of-jboss-home-name.adoc[leveloffset=+1]
+
+[[start_with_a_clean_server_install]]
```
emmartins, line 41: 1) Wrap the standard dist instructions in a header like this:
https://github.com/wildfly/quickstart/blob/main/helloworld/README-source.adoc?plain=1#L27
2) Change all sections to be wrapped to use header ===, and includes of sections to wrap should offset+2

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+== Start with a Clean Server Install
+
+It is important to start with a clean version of {productName} before testing this quickstart. Make sure you unzip or install a fresh {productName} instance.
+
+// Back Up the {productName} Managed Domain Configuration
```
emmartins, line 46: I guess this should be Standalone instead of Managed Domain?

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+include::../shared-doc/back-up-server-standalone-configuration.adoc[leveloffset=+1]
+// Start the {productName} Managed Domain
```
emmartins, line 48: Same as above?

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+include::../shared-doc/start-the-standalone-server.adoc[leveloffset=+1]
+
```
emmartins, line 50: It is missing the step to configure the server, ie run the script. You can copy and adapt from https://github.com/wildfly/quickstart/blob/main/ejb-security-context-propagation/README-source.adoc?plain=1#L83

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+// Build and Deploy the Quickstart
```
emmartins, line 51: This should be after the section wrt config changes

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+include::../shared-doc/build-and-deploy-the-quickstart.adoc[leveloffset=+1]
+
+[[review_the_modified_server_configuration]]
+== Review the Modified Server Configuration
+
+There are too many additions to the configuration file to list here. Feel free to compare the `standalone.xml` to the backup copy to see the changes made to configure the server to run this quickstart.
```
emmartins, line 57: This should describe the configuration changes, the ejb-multi-server lazy choice is not acceptable for a new quickstart :)

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+Similar result is for http://localhost:8080/dynamic-client-ssl-context/rest/port10443request:
+
+[source,options="nowrap"]
+----
+HTTP status of result is 200
+----
+
+// Server Distribution Testing
+:integrationTestsDirectory: app-web/src/test/
```
emmartins, line 89: please remove this attribute set, that is specific to ejb-multi-server multi-module design

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+include::../shared-doc/run-integration-tests-with-server-distribution.adoc[leveloffset=+1]
+
+// Undeploy the Quickstart
+include::../shared-doc/undeploy-the-quickstart.adoc[leveloffset=+1]
+
+// Restore the {productName} Managed Domain Configuration Manually
```
emmartins, line 95: Standalone?

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+include::../shared-doc/restore-standalone-server-configuration-manual.adoc[leveloffset=+1]
+
```
emmartins, line 97: This quickstart includes provisioned-server maven profile, but no instructions related with that, please add the following include to fix this:

include::../shared-doc/build-and-run-the-quickstart-with-provisioned-server.adoc[leveloffset=+1]

dynamic-client-ssl-context/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+        <artifactId>wildfly-quickstart-parent</artifactId>
+        <!--
+        Maintain separation between the artifact id and the version to help prevent
+        merge conflicts between commits changing the GA and those changing the V.
+        -->
+        <version>12</version>
+        <relativePath/>
+    </parent>
+
+    <properties>
```
emmartins, line 21: please resync property versions with values currently used by main branch

dynamic-client-ssl-context/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+    <name>Quickstart: dynamic-ssl-context</name>
+    <description>Dynamic client SSL context</description>
+
+    <licenses>
+        <license>
+            <name>Apache License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.html</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
```
emmartins, line 43: Please move this to below `<parent/>`

emmartins (changes requested):
This is also missing CI testing, but the whole CI is under redesign so I can add it this QS later to it, please note this means reviewers need to manually test the quickstart.

dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
kstekovi, line 95: **Build blocker:** `reload` breaks the `provisioned-server` build. When the WildFly Maven Plugin runs CLI scripts against its embedded server during `mvn package`, the `reload` command destroys the embedded process handle, resulting in:

```
Cannot invoke "BootstrappedEmbeddedProcess.getProcessStateNotifier()" because "this.embeddedProcess" is null
```

This was verified locally — the build fails after all CLI steps succeed. The fix is to comment it out, same as `helloworld-mutual-ssl/configure-ssl.cli` does:

```
#reload
```

The embedded packaging server does not need a reload — it re-provisions the server from scratch.

dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
```diff
@@ -0,0 +1,95 @@
...
+#
+
+# Script to configure dynamic SSL context in the application server
+
+# Create the server's keystores
+/subsystem=elytron/key-store=twoWayKS1:add(path=server1.keystore.P12,relative-to=jboss.server.config.dir,credential-reference={clear-text=secret},type=PKCS12)
+/subsystem=elytron/key-store=twoWayKS2:add(path=server2.keystore.P12,relative-to=jboss.server.config.dir,credential-reference={clear-text=secret},type=PKCS12)
+
+# Generate new server's key pairs. We'll use an DSA key of size 1024 and we'll use CN=localhost1 and CN=localhost2
+/subsystem=elytron/key-store=twoWayKS1:generate-key-pair(alias=localhost1 , algorithm=DSA, key-size=1024, validity=365, credential-reference={clear-text=secret}, distinguished-name="cn=localhost1")
```
kstekovi, line 13: DSA-1024 is deprecated and considered weak. This is a security quickstart — the key algorithm sends a message to users. Please use RSA-2048 or EC (e.g. `algorithm=RSA, key-size=2048`) for both server and client key pairs.

dynamic-client-ssl-context/configure-dynamic-client-ssl-context.cli
```diff
@@ -0,0 +1,95 @@
...
+/subsystem=elytron/key-store=twoWayTS2:add(path=server2.truststore.P12,relative-to=jboss.server.config.dir,credential-reference={clear-text=secret},type=PKCS12)
+
+# Set up the trust manager and keystore configuration for server's keystores
+/subsystem=elytron/key-manager=twoWayKM1:add(key-store=twoWayKS1,credential-reference={clear-text=secret})
+/subsystem=elytron/trust-manager=twoWayTM1:add(key-store=twoWayTS1)
+/subsystem=elytron/key-manager=twoWayKM2:add(key-store=twoWayKS2,credential-reference={clear-text=secret})
+/subsystem=elytron/trust-manager=twoWayTM2:add(key-store=twoWayTS2)
+
+# Create the server-ssl-context to reference the newly created trust-managers with required client auth option
+/subsystem=elytron/server-ssl-context=twoWaySSC1:add(key-manager=twoWayKM1,protocols=["TLSv1.2"],trust-manager=twoWayTM1,need-client-auth=true)
```
kstekovi, line 31: `protocols=["TLSv1.2"]` excludes TLS 1.3. Either add `"TLSv1.3"` to the list or omit the attribute entirely to inherit the server defaults. Same applies to lines 32, 75, and 76.

dynamic-client-ssl-context/src/main/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/Endpoint.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+@Path("/")
+public class Endpoint {
+
+    static final String DEFAULT_SERVER_HOST_FIRST_PORT = "https://localhost:9443";
+    static final String DEFAULT_SERVER_HOST_SECOND_PORT = "https://localhost:10443";
+    private static final String ARTIFACT_ID = "/dynamic-client-ssl-context/";
+
+    private ResteasyClientBuilder builder = (ResteasyClientBuilder) ClientBuilder.newBuilder();
+    private ResteasyClient client = builder.hostnameVerifier((s, sslSession) -> true).sslContext(SSLContext.getDefault()).build(); // resteasy client must set default ssl context otherwise it uses null
```
kstekovi, line 33: Two issues here:

1. **Resource leak:** `ResteasyClient` is a field on a request-scoped JAX-RS resource, so a new client (with its underlying connection pool) is created on every HTTP request and never closed.

2. **Unnecessary RESTEasy-specific API:** Both `.hostnameVerifier()` and `.sslContext()` are available on the standard `jakarta.ws.rs.client.ClientBuilder` since Jakarta REST 2.1 — there is no need to cast to `ResteasyClientBuilder` or reference RESTEasy types at all.

Suggested fix — use standard API and create the client only once:

```java
private static final Client CLIENT;
static {
    try {
        CLIENT = ClientBuilder.newBuilder()
            .hostnameVerifier((s, session) -> true)
            .sslContext(SSLContext.getDefault())
            .build();
    } catch (NoSuchAlgorithmException e) {
        throw new ExceptionInInitializerError(e);
    }
}
```

This also lets you drop the `resteasy-client-api` compile dependency from `pom.xml` entirely.

dynamic-client-ssl-context/pom.xml
kstekovi, line 106: `resteasy-client-api` at `compile` scope with a hardcoded version (`7.0.2.Final`) bypasses BOM management and bundles a RESTEasy jar into the WAR unnecessarily — the server already provides RESTEasy.

Since the only methods used (`hostnameVerifier`, `sslContext`) are part of the standard `jakarta.ws.rs.client.ClientBuilder` API (see comment on `Endpoint.java`), this dependency can be **removed entirely**. The standard Jakarta REST API is already available as `provided` via the EE BOM.

dynamic-client-ssl-context/pom.xml
```diff
@@ -0,0 +1,209 @@
...
+        <version.server>41.0.0.Beta1</version.server>
+        <!-- the versions for BOMs, Packs and Plugins -->
+        <version.bom.ee>${version.server}</version.bom.ee>
+        <version.plugin.wildfly>6.0.0.Final</version.plugin.wildfly>
+    </properties>
+
+    <artifactId>dynamic-client-ssl-context</artifactId>
+    <version>41.0.0.Final-SNAPSHOT</version>
+    <packaging>war</packaging>
+    <name>Quickstart: dynamic-ssl-context</name>
```
kstekovi, line 34: `<name>Quickstart: dynamic-ssl-context</name>` does not match the `artifactId` `dynamic-client-ssl-context`. Should be `Quickstart: dynamic-client-ssl-context`.

dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/BasicRuntimeIT.java
```diff
@@ -0,0 +1,134 @@
...
+public class BasicRuntimeIT {
+
+    private static final String DEFAULT_SERVER_DIR = System.getProperty("user.dir") + "/target/server";
+
+    @Test
+    public void testFirstHTTPSEndpointIsAvailable() throws IOException, URISyntaxException, KeyStoreException {
+        String serverConfigDir = getServerDir();
+
+        HttpGet request = new HttpGet(new URI(getServerHost(DEFAULT_SERVER_HOST_FIRST_PORT, 9443)));
+        KeyStore trustStore = loadKeyPairFromKeyStore(
```
kstekovi, line 50: The `trustStore` returned by `loadKeyPairFromKeyStore()` is assigned but never used — `getHttpClientWithSSL()` loads its own truststore independently directly from the file path. This is dead code and possibly a logic error (the constructed truststore is thrown away). Same issue on line 72. Consider removing these calls or using the returned `KeyStore` instead of re-loading from file.

dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/BasicRuntimeIT.java
```diff
@@ -0,0 +1,134 @@
...
+            throw new RuntimeException("Creating HttpClient with customized SSL failed.", e);
+        }
+    }
+
+    private String getServerDir() {
+        String serverDir  = System.getProperty("server.dir");
+        if (serverDir == null) {
+            serverDir = DEFAULT_SERVER_DIR;
+        }
+        return serverDir + "/standalone/configuration";
```
kstekovi, line 132: `getServerDir()` unconditionally appends `/standalone/configuration`. With the `provisioned-server` Maven profile, the server is provisioned under `target/server/`, not a standard installation layout, so `target/server/standalone/configuration` may not exist. This path assumption should be validated or made configurable via the `server.dir` property.

dynamic-client-ssl-context/src/test/java/org/jboss/as/quickstarts/dynamic_client_ssl_context/KeyStoreUtils.java
```diff
@@ -0,0 +1,45 @@
...
+    public static KeyStore loadKeyPairFromKeyStore(String serverDir, String keyStoreFile, String storePassword, String keyAlias, String keyStoreType) throws KeyStoreException {
+        FileInputStream stream = findFile(serverDir + "/" + keyStoreFile);
+        try {
+            KeyStore keyStore = KeyStore.getInstance(keyStoreType);
+            keyStore.load(stream, storePassword.toCharArray());
+            Certificate cert = keyStore.getCertificate(keyAlias);
+            KeyStore trustStore = keyStore.getInstance(keyStoreType);
+            trustStore.load(null, null);
+            trustStore.setCertificateEntry("server", cert);
+            trustStore.store(new FileOutputStream(serverDir + "/" + "client.truststore"), storePassword.toCharArray());
```
kstekovi, line 31: `loadKeyPairFromKeyStore()` writes a file `client.truststore` to disk as an undocumented side effect. This is unexpected from a utility method named "load". The file is also never read back anywhere in the test code, so the write serves no purpose. Either remove the write or move it to a dedicated method with a clear name.

dynamic-client-ssl-context/src/main/webapp/index.html
```diff
@@ -0,0 +1,20 @@
+<!--
+  ~ Copyright The WildFly Authors
+  ~ SPDX-License-Identifier: Apache-2.0
+  -->
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
+<html lang="en">
+<head>
+    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8; URL=rest/HelloWorld">
```
kstekovi, line 8: `URL=rest/HelloWorld` in the `<meta>` tag is leftover from a helloworld template and is not applicable here. Should be removed or updated to a relevant URL.

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
+include::../shared-doc/attributes.adoc[]
+
+= dynamic-client-ssl-context: Dynamic client SSL Context between Server and Client
+:author: Richard Záň
+:level: Intermidiate
```
kstekovi, line 5: Typo: `:level: Intermidiate` → `:level: Intermediate`

dynamic-client-ssl-context/README-source.adoc
```diff
@@ -0,0 +1,99 @@
...
+
+[source,options="nowrap"]
+----
+Click below to send request to different ports:
+
+Send request to port 9443
+Send request to port 10443
+----
+
+Each of links executes https GET request on server under specified ports. Each port uses a different TLS certificate. Both GET requests use the same dynamic client SSL context. For http://localhost:8080/dynamic-client-ssl-context/rest/port9443request, it will display:
```
kstekovi, line 74: The URL uses `http://` but the application only exposes HTTPS listeners (ports 9443 and 10443) — there is no plain HTTP listener on 8080 for this endpoint. Should be `https://`. Same issue on line 81.

dynamic-client-ssl-context/README-source.adoc
kstekovi, line 89: `:integrationTestsDirectory: app-web/src/test/` is incorrect — the integration tests are under `src/test/`, not `app-web/src/test/`. This will render a wrong path in the generated documentation. Should be removed (let the shared-doc default apply) or set to `src/test/`.

kstekovi (changes requested):
Reviewed the `dynamic-client-ssl-context` quickstart. The overall concept is sound — using a JAX-RS endpoint that calls itself on two different HTTPS ports is a clean way to demonstrate dynamic client SSL context selection. A few issues need to be addressed before this can merge, most importantly a build blocker in the CLI script.

emmartins:
@fRandOmizer any update on this?

fRandOmizer:
@emmartins I will continue work on this probably at the end of the September. Now I have more pressing matters at work with hard deadlines. Sorry for delay.

PR #1163, opened 2026-05-13
[WFLY-21912] Rework todo-backend README due to Glow, and due to similar content fo…

emmartins:
…r both wfly and eap
Issue: https://redhat.atlassian.net/browse/WFLY-21912


todo-backend/helm-install-prerequisites.adoc
```diff
@@ -1,19 +1,47 @@
...
-So we need to update the dependencies of our Helm Chart.
+To update the dependencies of our Helm Chart execute the following command.
 
 [source,options="nowrap",subs="+attributes"]
 ----
-$ helm dependency update charts/
+$ helm dependency update charts
+----
+
+= Environment variables for PostgreSQL
```
kstekovi, line 26: The section heading uses a level-0 title (`= Environment variables for PostgreSQL`), but the original in `additional-readme-cloud.adoc` used `===` (level 3 subsection). When this file is included inside the cloud deployment sections (with `leveloffset` applied by the shared docs), this will render at an incorrect level in the output. Please change it to `===`.

todo-backend/README-source.adoc
```diff
@@ -212,23 +147,37 @@ Please note that the quickstart includes integration tests, which may be execute
...
 //===========================================================
 // Openshift - START
 == Run the Backend on OpenShift
 
 // OpenShift
+:additional-readme-openshift: true
+:helm-install-prerequisites-openshift: ../todo-backend/helm-install-prerequisites.adoc
 include::../shared-doc/build-and-run-the-quickstart-with-openshift.adoc[leveloffset=+1]
-include::../todo-backend/additional-readme-openshift.adoc[leveloffset=+1]
+:!additional-readme-openshift:
```
kstekovi, line 168: `:additional-readme-openshift:` is correctly unset after the include, but `helm-install-prerequisites-openshift` (set on the line above) is never unset. This leaks the attribute into subsequent sections. Please add `:!helm-install-prerequisites-openshift:` here (and similarly `:!helm-install-prerequisites-kubernetes:` after the Kubernetes section) to mirror the handling of the other attributes.

todo-backend/additional-readme-cloud.adoc
```diff
@@ -1,70 +0,0 @@
...
-            configMapKeyRef:
-              name: postgres-configmap
-              key: POSTGRES_PASSWORD
-----
-
-When the application is deployed, the value for the `POSTGRESQL_PASSWORD` will be taken from the key `POSTGRES_PASSWORD`
-in the secret `postgres-configmap`.
-
-ifdef::additional-readme-openshift[]
-== Use the todobackend Web Frontend
```
kstekovi, line 34: This "Use the todobackend Web Frontend" section contained important practical guidance for OpenShift users — specifically how to retrieve the deployed route URL and construct the full path to use with the todobackend.com specs and client:

```bash
$ oc get route todo-backend -o jsonpath="{.spec.host}"
```

The todobackend.com links have been moved to the local testing section, but this cloud-specific URL discovery step is now missing. Without it, OpenShift users won't know how to find the URL to point todobackend.com at their deployed app. Could you add this back somewhere in the OpenShift section?

todo-backend/helm-install-prerequisites.adoc
```diff
@@ -1,19 +1,47 @@
...
+In local deployment the credentials were passed directly as the values of the environment variables.
+
+For {additional-readme-cloud-platform}, we rely on secrets so that the credentials are never copied outside {additional-readme-cloud-platform}:
+
+[source,options="nowrap"]
+----
+deploy:
+  env:
+    - name: POSTGRESQL_PASSWORD
+        - name: POSTGRESQL_DATABASE
```
kstekovi: Minor/pre-existing: `- name: POSTGRESQL_DATABASE` is indented as a sub-item of `- name: POSTGRESQL_PASSWORD` rather than at the same level. Since this block is being moved and touched anyway, it would be a good opportunity to fix the indentation.

kstekovi (changes requested):
Thanks for the cleanup! Reducing the ifdef/ifndef duplication and reflecting the Glow-based provisioning is a clear improvement. I have a couple of issues to flag before this is ready to merge.

todo-backend/helm-install-prerequisites.adoc
emmartins: This should actually be good now, and before was wrong, we should always start adocs with level 1, and use the level-offset, when including the file, as its now done on https://github.com/emmartins/quickstart/blob/fix-eap-todo-os-instructions/shared-doc/build-and-run-the-quickstart-with-openshift.adoc?plain=1#L17

todo-backend/README-source.adoc
emmartins: Fixed

todo-backend/additional-readme-cloud.adoc
emmartins: Those instructions where duped, they are already provided by content coming from https://github.com/emmartins/quickstart/blob/fix-eap-todo-os-instructions/todo-backend/README-source.adoc?plain=1#L167C34-L167C37

todo-backend/helm-install-prerequisites.adoc
emmartins: Fixed

emmartins:
Thanks for you review @kstekovi , changes applied, please re-review.

todo-backend/helm-install-prerequisites.adoc
```diff
@@ -1,19 +1,46 @@
...
 ----
-$ helm dependency update charts/
+$ helm dependency update charts
+----
+
+= Environment variables for PostgreSQL
+
+The Helm Chart also contains the environment variables required to connect to the PostgreSQL database.
+
+In local deployment the credentials were passed directly as the values of the environment variables.
```
kstekovi: Nit: this sentence refers to "local deployment" but this file is only ever included inside the OpenShift and Kubernetes sections. It reads a bit oddly in that context. Consider removing it or rephrasing, e.g. _"Unlike local deployment where credentials are set directly, for {additional-readme-cloud-platform} we rely on secrets…"_

todo-backend/helm-install-prerequisites.adoc
```diff
@@ -1,19 +1,46 @@
...
+[source,options="nowrap"]
+----
+    deploy:
+      env:
+        - name: POSTGRESQL_PASSWORD
+          valueFrom:
+            configMapKeyRef:
+              name: postgres-configmap
+              key: POSTGRES_PASSWORD
 ----
```
kstekovi, line 41: Question (no change requested): the example only shows `POSTGRESQL_PASSWORD`, but `charts/values.yaml` also configures `POSTGRESQL_DATABASE` and `POSTGRESQL_USER` from the same configmap. Is this intentional — i.e. the snippet is meant to illustrate the pattern rather than be a complete reference — or should the other two variables be included as well?

kstekovi (commented):
Re-review after verifying the generated HTML output locally. All previous blocking comments have been addressed. Two minor items remain — one nit and one clarifying question.

todo-backend/helm-install-prerequisites.adoc
emmartins: Yes this was intended.

todo-backend/helm-install-prerequisites.adoc
emmartins: Will do suggested change, thanks.

emmartins: Done.

kstekovi:
Thanks for updates. LGTM.

PR #1169, opened 2026-05-20
Quickstart for MCP server

jmesnil:


jmesnil:
@ehsavoie, this would be a Quickstart after the MCP server features are promoted to stability level (tracked by https://github.com/wildfly/wildfly-proposals/pull/830).
It is based on your wildfly-weather :)

PR #1171, opened 2026-05-21
[WFLY-21323] http-custom-mechanism is accepting the req with invalid password

rafaeldarosagit:
Issue: https://redhat.atlassian.net/browse/WFLY-21323
Description: Authentication mechanism was accepting requests with invalid passwords when users had valid authorization groups. Added missing return statement after authentication failed.

### DIFF
--- a/http-custom-mechanism/webapp/src/test/java/org/jboss/as/quickstart/http_custom_mechanism/BasicRuntimeIT.java
+++ b/http-custom-mechanism/webapp/src/test/java/org/jboss/as/quickstart/http_custom_mechanism/BasicRuntimeIT.java
@@ -15,6 +15,7 @@
  */
 package org.jboss.as.quickstart.http_custom_mechanism;
 
+import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


emmartins:
thanks @rafaeldarosagit , I will review this week

http-custom-mechanism/webapp/src/test/java/org/jboss/as/quickstart/http_custom_mechanism/BasicRuntimeIT.java
emmartins: Password here is wrong too, it should be `quickstartPwd1!`

emmartins (changes requested):
Please also squash the commits into a single one, and which commit message should start with `[WFLY-21323] `

http-custom-mechanism/webapp/src/test/java/org/jboss/as/quickstart/http_custom_mechanism/BasicRuntimeIT.java
rafaeldarosagit: @emmartins done! Thanks!

emmartins (approved):

PR #1175, opened 2026-06-03
[WFLY-21115] Sync MP RP Kafka QS code and README

kabir:
https://redhat.atlassian.net/browse/WFLY-21115

### DIFF
--- a/microprofile-reactive-messaging-kafka/README-source.adoc
+++ b/microprofile-reactive-messaging-kafka/README-source.adoc
@@ -400,7 +400,6 @@ Additionally we want to set a Kafka key for the entries so that we can take adva
     @Incoming("sender")
     @Outgoing("to-kafka")
     public Message<TimedEntry> sendToKafka(String msg) {
-        // Simpler example for debugging
         TimedEntry te = new TimedEntry(new Timestamp(System.currentTimeMillis()), msg);
         Message<TimedEntry> m = Message.of(te);
         // Just use the hash as the Kafka key for this example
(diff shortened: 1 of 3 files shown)


emmartins:
@kstekovi sorry I went too fast and merged this, and now noticed you were the reporter, please review and if needed we will add another PR

kstekovi:
Thanks for the cleanup in this PR! While reviewing the changes, I noticed a few things worth discussing:

**1. Unchecked conversion warning (confirmed)**
Removing `@SuppressWarnings("unchecked")` from `receiveFromKafka` introduced a compiler warning:
```
[WARNING] MessagingBean.java:[89,69] unchecked conversion
  required: IncomingKafkaRecordMetadata<Integer, TimedEntry>
  found:    IncomingKafkaRecordMetadata
```
`KafkaMetadataUtil.readIncomingKafkaMetadata()` returns a raw `Optional<IncomingKafkaRecordMetadata>`, so the annotation was there intentionally. Was its removal deliberate?

**2. Unguarded `.get()` on Optional**
On the same line, `KafkaMetadataUtil.readIncomingKafkaMetadata(message).get()` calls `.get()` without checking presence. Previously this was masked by the `try/catch`. If Kafka metadata is ever absent, this will throw a `NoSuchElementException`. Is this expected to always be present in this context? If so, `.orElseThrow()` would make that assumption explicit.

**3. Quoted `group.id` value**
The new config property added to the README:
```properties
mp.messaging.connector.smallrye-kafka.group.id="microprofile-reactive-messaging-kafka-group-id"
```
In `.properties` files, values are not quoted — the double quotes become part of the string, so the actual `group.id` seen by Kafka would include the quotes. Is this intentional, or should the quotes be removed?

kabir:
@kstekovi @emmartins Follow-up to address the comments in #1177

PR #1177, opened 2026-06-17
[WFLY-21115] MP RM Kafka QS Code/Doc sync follow-up

kabir:
https://redhat.atlassian.net/browse/WFLY-21115

Follows up on #1175


kstekovi:
Thanks for the quick follow-up! All three issues have been addressed:

- **`@SuppressWarnings("unchecked")`** restored on `receiveFromKafka` — fixes the compiler warning.
- **`.get()` replaced with `.orElseThrow()`** — makes the assumption explicit and gives a clearer exception if metadata is ever absent.
- **Quoted `group.id` fixed** in both `microprofile-config.properties` and both README files, and the warning message example in the README is also updated to reflect the unquoted value.

LGTM, the changes look correct.

emmartins (approved):

PR #1179, opened 2026-06-19
[WFLY-20952] Added integration tests to the JTS quickstart

jmfinelli:
Addresses [WFLY-20952](https://redhat.atlassian.net/browse/WFLY-20952)


jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
```diff
@@ -0,0 +1,195 @@
...
+                        "addCustomerForm:name", name,
+                        "addCustomerForm:add", "Add",
+                        "jakarta.faces.ViewState", viewState)))
+                .build();
+        return client.send(postRequest, HttpResponse.BodyHandlers.ofString());
+    }
+
+    private long getMessagesAdded() throws IOException {
+        try (ModelControllerClient client = ModelControllerClient.Factory.create(
+                InetAddress.getByName("localhost"), SERVER2_MGMT_PORT)) {
```
emmartins: This should use the host in getServerHost().

jts/README-source.adoc
```diff
@@ -231,6 +231,15 @@ INFO  [class org.jboss.as.quickstarts.cmt.jts.mdb.HelloWorldMDB] (Thread-97 (Act
 
 The web page will also change and show you the new list of customers.
 
+NOTE: You can also verify the state of the JMS queue on application server 2 by using the management CLI. Connect to the second server's management interface and run the following command, replacing `__DEPLOYMENT_NAME__` with the actual deployment file name (e.g. `{artifactId}-application-component-2-{version}.jar`):
```
emmartins: {version} is not defined, so you either pass it from Maven as done at https://github.com/jmfinelli/quickstart-wildfly/blob/WFLY-20952/microprofile-fault-tolerance/pom.xml#L147 , or you set the deployment name's to have no version, as done for application-component-1 at https://github.com/jmfinelli/quickstart-wildfly/blob/WFLY-20952/jts/application-component-1/pom.xml#L109

jts/README-source.adoc
```diff
@@ -231,6 +231,15 @@ INFO  [class org.jboss.as.quickstarts.cmt.jts.mdb.HelloWorldMDB] (Thread-97 (Act
...
+
+[source,options="nowrap"]
+----
+/deployment=__DEPLOYMENT_NAME__/subsystem=messaging-activemq/server=default/jms-queue=jts-quickstart:read-attribute(name=messages-added)
+----
+
+This shows the total number of messages added to the queue since the server started. You can find the deployment name by running `ls /deployment` in the CLI.
+
 // Undeploy the Quickstart
```
emmartins, line 258: We need testing instructions, usually this is done by adding the following include, before the undeploy one:

```
// Server Distribution Testing
include::../shared-doc/run-integration-tests-with-server-distribution.adoc[leveloffset=+1]
```

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
```diff
@@ -0,0 +1,195 @@
...
+                        "addCustomerForm", "addCustomerForm",
+                        "addCustomerForm:name", name,
+                        "addCustomerForm:add", "Add",
+                        "jakarta.faces.ViewState", viewState)))
+                .build();
+        return client.send(postRequest, HttpResponse.BodyHandlers.ofString());
+    }
+
+    private long getMessagesAdded() throws IOException {
+        try (ModelControllerClient client = ModelControllerClient.Factory.create(
```
emmartins: You could use http client for this too, and don't depend on WildFly Core... Honestly I am not sure if the Core's ModelControllerClient should be used by user apps.

emmartins (changes requested):

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
```diff
@@ -0,0 +1,195 @@
...
+                InetAddress.getByName("localhost"), SERVER2_MGMT_PORT)) {
+            String deploymentName = findComponent2Deployment(client);
+            ModelNode op = new ModelNode();
```
kstekovi: `ModelControllerClient.execute()` never throws on a failed DMR operation — it always returns a `ModelNode` with `outcome=failed` and no `result` key. Calling `.asLong()` (or `.asList()`) on the resulting UNDEFINED node throws `IllegalArgumentException`, which surfaces as a confusing internal error rather than a meaningful test failure message.

Suggest checking the outcome before reading the result in both `getMessagesAdded()` and `findComponent2Deployment()`. The needed constants (`OUTCOME`, `SUCCESS`) are already available from `ClientConstants`:

```java
// add to imports
import static org.jboss.as.controller.client.helpers.ClientConstants.OUTCOME;
import static org.jboss.as.controller.client.helpers.ClientConstants.SUCCESS;

private long getMessagesAdded() throws IOException {
    try (ModelControllerClient client = ModelControllerClient.Factory.create(
            InetAddress.getByName("localhost"), SERVER2_MGMT_PORT)) {
        String deploymentName = findComponent2Deployment(client);
        ModelNode op = new ModelNode();
        op.get(OP).set(READ_ATTRIBUTE_OPERATION);
        op.get(OP_ADDR).add("deployment", deploymentName)
                       .add("subsystem", "messaging-activemq")
                       .add("server", "default")
                       .add("jms-queue", "jts-quickstart");
        op.get(NAME).set("messages-added");
        ModelNode result = client.execute(op);
        if (!SUCCESS.equals(result.get(OUTCOME).asString())) {
            throw new RuntimeException("Failed to read messages-added: "
                    + result.get("failure-description").asString());
        }
        return result.get(RESULT).asLong();
    }
}

private String findComponent2Deployment(ModelControllerClient client) throws IOException {
    ModelNode op = new ModelNode();
    op.get(OP).set(READ_CHILDREN_NAMES_OPERATION);
    op.get(CHILD_TYPE).set("deployment");
    ModelNode result = client.execute(op);
    if (!SUCCESS.equals(result.get(OUTCOME).asString())) {
        throw new RuntimeException("Failed to list deployments on Server 2: "
                + result.get("failure-description").asString());
    }
    for (ModelNode entry : result.get(RESULT).asList()) {
        String name = entry.asString();
        if (name.startsWith("jts-application-component-2")) {
            return name;
        }
    }
    throw new RuntimeException("jts-application-component-2 deployment not found on Server 2");
}
```

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    @Test
+    public void testAddCustomer() throws Exception {
+        HttpClient client = createHttpClient();
+        String uniqueName = "TestCustomer_" + System.currentTimeMillis();
+        long baselineMessages = getMessagesAdded();
+
+        HttpResponse<String> response = submitCustomerForm(client, uniqueName);
+        assertEquals(200, response.statusCode());
```
kstekovi, line 70: `Thread.sleep(2000)` is too fragile for a JTS scenario on shared/overloaded CI infrastructure. The delay covers a JTS two-phase commit across two separate server instances plus the stat propagation to Server 2's management API — under load this can exceed 2 seconds, leading to potential intermittent failures.

Two helpers handle both directions reliably:

```java
private void waitForMessagesAdded(long expected, long timeoutMs)
        throws IOException, InterruptedException {
    long deadline = System.currentTimeMillis() + timeoutMs;
    while (System.currentTimeMillis() < deadline) {
        if (getMessagesAdded() == expected) return;
        Thread.sleep(500);
    }
    assertEquals("Timed out waiting for messages-added to reach " + expected,
            expected, getMessagesAdded());
}

private void assertMessagesStayAt(long expected, long durationMs)
        throws IOException, InterruptedException {
    long deadline = System.currentTimeMillis() + durationMs;
    while (System.currentTimeMillis() < deadline) {
        assertEquals(expected, getMessagesAdded());
        Thread.sleep(500);
    }
}
```

In `testAddCustomer`:
```java
waitForMessagesAdded(baselineMessages + 1, 30_000);
```

In `testDuplicateCustomerRollback`, capture the baseline before the first submission:
```java
long baselineMessages = getMessagesAdded();

HttpResponse<String> response = submitCustomerForm(client, uniqueName);
// ... existing assertions ...
waitForMessagesAdded(baselineMessages + 1, 30_000);

response = submitCustomerForm(client, uniqueName);    // submit duplicate
// ... existing assertions ...
// On overloaded CI infrastructure a rolled-back JTS transaction may take several seconds
// to fully resolve across both servers — poll for the full duration to ensure no late
// message arrives from the rolled-back transaction
assertMessagesStayAt(baselineMessages + 1, 10_000);
```

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
kstekovi: Agreed with @emmartins. The same applies to `SERVER2_MGMT_PORT` — the management port is also hardcoded to `10090` with no override, creating the same problem.

The `ejb-txn-remote-call` quickstart shows the established pattern for a second server — each server gets its own configurable host with a sensible default ([server module](https://github.com/wildfly/quickstart/blob/main/ejb-txn-remote-call/server/src/test/java/org/jboss/as/quickstarts/ejb/BasicRuntimeIT.java)). Suggest applying the same approach here for both the management host and port:

```java
private static final String DEFAULT_SERVER2_MGMT_HOST = "localhost";
private static final int DEFAULT_SERVER2_MGMT_PORT = 10090;

private String getServer2MgmtHost() {
    String host = System.getenv("SERVER2_MGMT_HOST");
    if (host == null) {
        host = System.getProperty("server2.mgmt.host");
    }
    if (host == null) {
        host = DEFAULT_SERVER2_MGMT_HOST;
    }
    return host;
}

private int getServer2MgmtPort() {
    String port = System.getProperty("server2.mgmt.port");
    return port != null ? Integer.parseInt(port) : DEFAULT_SERVER2_MGMT_PORT;
}
```

Then in `getMessagesAdded()`:
```java
ModelControllerClient.Factory.create(
    InetAddress.getByName(getServer2MgmtHost()), getServer2MgmtPort())
```

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
```diff
@@ -0,0 +1,195 @@
...
+        final HttpClient client = HttpClient.newBuilder()
+                .followRedirects(HttpClient.Redirect.ALWAYS)
+                .connectTimeout(Duration.ofMinutes(1))
+                .build();
+        final HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+        assertEquals(200, response.statusCode());
+        final String[] bodyLines = response.body().lines().toArray(String[]::new);
+        assertEquals("<meta http-equiv=\"Refresh\" content=\"0; URL=addCustomer.jsf\">", bodyLines[bodyLines.length - 3]);
+    }
+
```
kstekovi, line 62: This pattern (`bodyLines[bodyLines.length - 3]`) is consistent with other quickstarts in this repo (e.g. `ejb-multi-server`), so it works. However it is sensitive to any whitespace or line count change in the rendered HTML — a minor server version bump could break the assertion without any functional regression.

A `contains` check would be equally readable and immune to line count changes:

```java
assertTrue(response.body().contains(
    "<meta http-equiv=\"Refresh\" content=\"0; URL=addCustomer.jsf\">"));
```

kstekovi (commented):
Review of the integration tests added to the JTS quickstart.

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
jmfinelli: Thanks

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
jmfinelli: Thanks!

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
jmfinelli: Thanks! Hopefully, I interpreted your comment correctly. Let me know what you think.

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
jmfinelli: Thanks!

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
jmfinelli: This code has been deleted.

jts/README-source.adoc
jmfinelli: Done. Thanks!

jts/README-source.adoc
jmfinelli: Done. Thanks!

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
jmfinelli: Thanks!

jts/README-source.adoc
```diff
@@ -231,6 +231,17 @@ INFO  [class org.jboss.as.quickstarts.cmt.jts.mdb.HelloWorldMDB] (Thread-97 (Act
...
+
+[source,options="nowrap"]
+----
+/deployment={artifactId}-application-component-2.war/subsystem=messaging-activemq/server=default/jms-queue=jts-quickstart:read-attribute(name=messages-added)
+----
+
+This shows the total number of messages added to the queue since the server started.
+
+// Server Distribution Testing
+include::../shared-doc/run-integration-tests-with-server-distribution.adoc[leveloffset=+1]
```
kstekovi, line 242: When running the integration tests, the `testDuplicateCustomerRollback` test intentionally submits a duplicate customer name to verify the JTS transaction rollback. The following warnings and errors will appear in the server 1 log and are safe to ignore:

```
WARN  [org.hibernate.engine.jdbc.spi.SqlExceptionHelper] (default task-3) SQL Error: 0, SQLState: 23505
ERROR [org.hibernate.engine.jdbc.spi.SqlExceptionHelper] (default task-3) ERROR: duplicate key value violates unique constraint "uk6m4fuhxj4ckixi1vsb8osii7r"
WARN  [com.arjuna.ats.jts] (default task-3) ARJUNA022024: ArjunaTransactionImple.doBeforeCompletion caught exception: org.omg.CORBA.UNKNOWN: ...
WARN  [com.arjuna.ats.jtax] (default task-3) ARJUNA024068: Caught exception:: org.omg.CORBA.TRANSACTION_ROLLEDBACK: ...
ERROR [org.jboss.as.ejb3.invocation] (default task-3) WFLYEJB0034: Jakarta Enterprise Beans Invocation failed on component CustomerManagerEJB ...
WARNING [org.jboss.as.quickstarts.cmt.controller.CustomerManager] (default task-3) Caught a duplicate: ...
```

Suggest adding a NOTE block after the integration testing include, following the same pattern as the existing deploy warnings note above.

kstekovi (commented):
Apologies for the late comments — these should have been caught in the first review round but I ran out of time yesterday.

kstekovi:
Nitpick on `jts/README-source.adoc` (pre-existing, out of scope for this PR): the AsciiDoc build emits warnings at lines 157, 165, 172 — `list item index: expected N, got 1` — because the PostgreSQL driver setup section repeats `1.` for each numbered step instead of using explicit numbering (`1.`, `2.`, `3.`, `4.`). This has no effect on the rendered output; the HTML displays the correct sequence. The fix would be replacing the repeated `1.` with explicit indices.

jts/README-source.adoc
jmfinelli: Thanks! Done

kstekovi (approved):
All review comments from both rounds have been addressed. The integration tests are well structured, the timing issues are handled correctly with polling helpers, the management client dependency has been cleanly replaced with a REST endpoint, and the README now documents the expected warnings for both deployment and integration test phases. LGTM!

jts/README-source.adoc
emmartins: Please take the opportunity to fix this, this attr is not defined, replace with `urn:jboss:domain:iiop-openjdk:3.0`

emmartins: Please take the opportunity to fix this, this attr is not defined, replace with `urn:jboss:domain:transactions:7.0`

emmartins: Let's also take the opportunity to get ridden of this link to WFLY docs, not really needed. I suggest to replace both paragraphs with:

`First of all, a JDBC driver needs to be installed. The following command downloads the PostgreSQL driver automatically through Maven. When needed, replace __{jbossHomeName}_1__ with the path to your server
`

jts/README-source.adoc
```diff
@@ -231,6 +231,32 @@ INFO  [class org.jboss.as.quickstarts.cmt.jts.mdb.HelloWorldMDB] (Thread-97 (Act
 
 The web page will also change and show you the new list of customers.
 
+NOTE: You can also verify the state of the JMS queue on application server 2 by using the management CLI. Connect to the second server's management interface and run:
+
+[source,options="nowrap"]
```
emmartins: You need to activate attr substitutions here:
`[source, subs="attributes+", options="nowrap"]`

jts/README-source.adoc
emmartins: Please take the opportunity to fix this, this attr is not defined, replace with `urn:jboss:domain:iiop-openjdk:3.0`

emmartins: Please take the opportunity to fix this, this attr is not defined, replace with `urn:jboss:domain:transactions:7.0`

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.junit.Test;
+
+public class BasicRuntimeIT {
+
+    private static final String DEFAULT_SERVER_HOST = "http://localhost:8080";
+    private static final String APP_CONTEXT = "/jts-application-component-1";
+    private static final String DEFAULT_SERVER2_HOST = "http://localhost:8180";
```
emmartins: Just use the port offset to calculate this from getServerHost(), no need to have 2 params

emmartins (changes requested):
Sorry, we need one more round.

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
jmfinelli: Done. Thanks!

jts/README-source.adoc
jmfinelli: Done. Thanks!

jmfinelli: Done. Thanks!

jts/README-source.adoc
jmfinelli: Done. Thanks!

jts/README-source.adoc
jmfinelli: Done. Thanks!

jmfinelli: Done. Thanks!

jmfinelli: Done. Thanks!

emmartins (approved):

emmartins:
Thanks @jmfinelli

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
```diff
@@ -0,0 +1,186 @@
...
+
+    private String getServerHost(int portOffset) {
+        String host = System.getenv("SERVER_HOST");
+        if (host == null) {
+            host = System.getProperty("server.host");
+        }
+        if (host == null) {
+            host = DEFAULT_SERVER_HOST;
+        }
+        return host + ":" + (DEFAULT_SERVER_PORT + portOffset);
```
emmartins: Actually, the server_host args passed in CLI are supposed to have the port... So here, if offset is not zero, then you will need to parse it and add the port offset (if not present: 80 for http, 443 for https ) before returning it.

emmartins (changes requested):
Sorry missed something

jts/application-component-1/src/test/java/org/jboss/as/quickstarts/cmt/jts/BasicRuntimeIT.java
jmfinelli: Cool, that makes sense. Thanks Eduardo

jmfinelli: Done

emmartins (approved):

emmartins:
@jmfinelli this is the way, thank you

PR #1197, opened 2026-08-11
[WFLY-22003] Replace JPA initialization with JDBC CDI bean

Riovo:
Fixes: https://redhat.atlassian.net/browse/WFLY-22003

Replaces the current JPA-based database initialization with a CDI bean using direct JDBC. This eliminates the WildFly Glow unbound datasource warning that popped up because persistence.xml referenced a CLI-defined datasource.

The CDI bean observes application startup (@Observes @Initialized(ApplicationScoped.class)) to initialize the H2 database schema and test users for the Elytron JDBC realm.


servlet-security/src/main/java/org/jboss/as/quickstarts/servlet_security/DatabaseInitializer.java
```diff
@@ -0,0 +1,225 @@
...
+import jakarta.annotation.security.PermitAll;
+import jakarta.ejb.Singleton;
+import jakarta.ejb.Startup;
+
+/**
+ * Initializes the database schema and data for Elytron JDBC realm authentication.
+ * This singleton EJB runs at application startup to create tables and populate
+ * test user credentials if they don't already exist.
+ *
+ * @author Mohammed Abourass
```
luck3y: minor: usually add your email here like <foo@ibm.com> or whatever mail address you want to use.

servlet-security/src/main/java/org/jboss/as/quickstarts/servlet_security/DatabaseInitializer.java
```diff
@@ -0,0 +1,225 @@
...
+            LOGGER.log(Level.SEVERE, "Failed to initialize database", e);
+            throw new RuntimeException("Database initialization failed", e);
+        }
+    }
+
+    private void createTables() throws SQLException {
+        Connection conn = null;
+        Statement stmt = null;
+
+        try {
```
luck3y: One thing about quickstarts is that people will tend to copy your code. Perhaps consider using the try-with-resources style here vs an explict close in finally?

luck3y: this also applies to the other uses of try {} finally {} below

luck3y (approved):
LGTM, I added some suggestions, feel free to disagree :)

servlet-security/src/main/java/org/jboss/as/quickstarts/servlet_security/DatabaseInitializer.java
```diff
@@ -0,0 +1,225 @@
+/*
```
emmartins, line 1: Use the latest header:

```
/*
 */
```

emmartins (changes requested):

emmartins:
Personally I dislike the usage of an EJB for this, I know this was suggested in the original JIRA but EJB is kind of legacy and considered an expensive resource, I suggest instead either:

1) In case DB initialization separated from the Servlet is prefered, to turn DatabaseInitializer into a CDI bean that observes app initialization:

```
@ApplicationScoped
public class DatabaseInitializer {
    ...
    public void initializeDatabase(@Observes @Initialized(ApplicationScoped.class) Object init) {
       ...
    }
    ...
}
```

2) Use nothing else than the existing Servlet, overriding its init() method, where you will invoke DatabaseInitializer, and changing the WebServlet annotation to load-init on startup:

```
@WebServlet(value = "/SecuredServlet", loadOnStartup = 1)
@ServletSecurity(@HttpConstraint(rolesAllowed = { "quickstarts" }))
public class SecuredServlet extends HttpServlet {
    ...
    @Resource(lookup = "java:jboss/datasources/ServletSecurityDS")
    private DataSource dataSource;

    @Override
    public void init() throws ServletException {
        new DatabaseInitializer(dataSource).initializeDatabase();
    }
    ...
}

```

Either requires no changes to the CLI script, or aditional dependencies.

PS: Ken remarks would still need to be considered.

Riovo:
Thank you @emmartins and @luck3y for your comments. I have since implemented everything, please let me know if everything is good or if changes need to be done.

servlet-security/pom.xml
```diff
@@ -93,8 +93,8 @@
         </dependency>
 
         <dependency>
-            <groupId>jakarta.persistence</groupId>
-            <artifactId>jakarta.persistence-api</artifactId>
+            <groupId>jakarta.ejb</groupId>
```
emmartins: Please remove this dependency, no longer needed.

emmartins (changes requested):

servlet-security/pom.xml
Riovo: will do, thank you!

emmartins:
@Riovo please squash all commits into one too :)

Riovo:
> @Riovo please squash all commits into one too :)

Hehe will do, thank you for you time :)

Riovo:
@darranl

emmartins (approved):

emmartins:
LGTM @Riovo, I requested review from @kstekovi too, he was the original issue[1] reporter.

[1] https://redhat.atlassian.net/browse/JBEAP-33412

kstekovi:
LGTM overall, one non-blocking nit: `insertTestData()` runs each `INSERT` with auto-commit on, and `userExists()` only gates on the `USERS` row. If a statement fails mid-sequence (e.g. between the USERS and ROLES inserts), a later redeploy would see `quickstartUser` already present and skip re-inserting the missing `ROLES`/`USERS_ROLES` rows.

Low risk given the small dataset, but `conn.setAutoCommit(false)` + commit at the end / rollback on `SQLException` would make the whole insert block atomic and avoid that edge case. Up to you — fine to merge as-is either way.

emmartins:
> LGTM overall, one non-blocking nit: `insertTestData()` runs each `INSERT` with auto-commit on, and `userExists()` only gates on the `USERS` row. If a statement fails mid-sequence (e.g. between the USERS and ROLES inserts), a later redeploy would see `quickstartUser` already present and skip re-inserting the missing `ROLES`/`USERS_ROLES` rows.
> 
> Low risk given the small dataset, but `conn.setAutoCommit(false)` + commit at the end / rollback on `SQLException` would make the whole insert block atomic and avoid that edge case. Up to you — fine to merge as-is either way.

@kstekovi  Since the datasource uses default autocommit and the CDI Bean is not transactional, I think it will autocommit when every connection instance is closed, and if that is the case all needed to enforce your enhancement would be to create the connection at the event handler and, pass it along the 2 methods in it.

@Riovo can you please investigate and optimize this code if needed?

